PR #20, opened 2025-06-19
feat: Android library setup

ovitrif:
This PR introduces Android library setup to support declaring bitkit-core as dependency in android app projects.

### Updates
- Android library project scaffold
- Move generated android native libs and bindings to the nested dirs expected by the android library project
- JitPack distribution support
- Updates the android build generation script to support the new android library setup
- Android compatibility with 16 KB devices - see [android docs](https://developer.android.com/16kb-page-size)


ovitrif:
Tested successfully by importing as dependency via jitpack in bitkit-android PR [204](https://github.com/synonymdev/bitkit-android/pull/204) 🎉

coreyphillips:
Otherwise, LGTM!

PR #22, opened 2025-06-25
fix: update android build file

coreyphillips:
This PR:
- Removes newline entry generated from` build_android.sh`.


ovitrif (approved):
Awesome find & fix 🎉 👍🏻

PR #32, opened 2025-08-11
feat: github packages android distribution

ovitrif:
This PR:
- setups distribution for android via github packages
- adds gradle-publish workflow
- removes jitpack setup

Tested via bitkit-android PR:
- https://github.com/synonymdev/bitkit-android/pull/294


ovitrif:
@coreyphillips once merged, please create a new release so we can use this in Android 🙏🏻 

Easiest is to just increment to the next version (done automatically by `./build.sh all` ?!)

PR #33, opened 2025-08-11
chore: update bindings

coreyphillips:
- Updates bindings
- Upgrades version to 0.0.10


ovitrif:
The main purpose I asked for this new release was to be able to consume the github package for Android that was supposed to be created after the workflow from `gradle-publish.yml` gets triggered on new release.

But there is no github package, nor can I see see the `Actions` tab. Maybe they're disabled on the repo 🤷🏻, in any case will wait for access rights before I can dig further.

I also found another issue: can't use `GITHUB_TOKEN` as is because org policy disables `Read and write permissions` Workflow permissions. So I will have to set it up differently, using a PAT.
Will definitely need access rights to be able to investigate this further and debug until I can get it working.

Going to ask tomorrow again.

ovitrif (approved):
LGTM, can be merged actually

We can create a package for android at any time after merging the PR that allows to trigger the publishing separate from releases:
- #34

PR #45, opened 2025-10-29
Add closed channels table

ben-kaufman:
Adds a table for closed_channels as part of the activities DB.

The closed channel details saved are the following:
   ```
channel_id: String,
   counterparty_node_id: String,
   funding_txo_txid: String,
   funding_txo_index: u32,
   channel_value_sats: u64,
   closed_at: u64,
   outbound_capacity_msat: u64,
   inbound_capacity_msat: u64,
   counterparty_unspendable_punishment_reserve: u64,
   unspendable_punishment_reserve: u64,
   forwarding_fee_proportional_millionths: u32,
   forwarding_fee_base_msat: u32,
   channel_name: String,
   channel_closure_reason: String,
```

Also, a new index is added here for the new closed_channels table based on the funding tx id.

The new functionalities now exposed are:
- `insert_closed_channel`
- `get_closed_channel_by_id`
- `get_all_closed_channels`
- `remove_closed_channel_by_id`
- `wipe_all_closed_channels`

Also, the `wipe_all` function was updated to also delete all the closed channels.


ovitrif (commented):
Should we instead of `ClosedChannelDetails` record `ChannelDetails` model with `status = CLOSED` enum?

Wondering, design-wise, if this is perhaps more future-proof 🤔 

Just wanting your input and @coreyphillips 's; not saying we must.

ben-kaufman:
I think it makes sense to have this explicitly for closed channels since open channels will always be managed by the node, ie there shouldn't be any future scenario where we need to also separately track channels which are open since the ldk node must always manage these.

That said, if you think there could be a different reason now or in the future to have all channels details stored here we can make this change.

coreyphillips:
@ben-kaufman updated your role so you can create releases as needed.

PR #49, opened 2025-11-06
Add upsert tags and get all tags metadata

ben-kaufman:
This PR adds support for tag backup and restore with full activity metadata.

Changes:
- Add ActivityTagsMetadata struct with full activity metadata (id, payment_hash,
  tx_id, address, is_receive, tags, created_at) for backup format
- Add ActivityTags struct (activity_id, tags) for simple tag upsert operations
- Implement get_all_tag_metadata() method that joins with activities and
  onchain_activity tables to return full metadata
- Update upsert_tags() to accept ActivityTags (simplified structure) using
  INSERT OR IGNORE for additive behavior

New calls exposed:
- Backup: get_all_tag_metadata() returns Vec<ActivityTagsMetadata> with full
  activity context.
- Restore: upsert_tags() accepts Vec<ActivityTags> with just activity_id and
  tags.


ovitrif (approved):
LGTM 🎉 

Thank you for adding this implementation so fast 🙏🏻

PR #51, opened 2025-11-12
Feat: Add pre activity metadata

ben-kaufman:
This PR adds pre-activity metadata to the activities DB.

The goal of the pre-activity metadata is to keep data relevant to activities before activities are actually created.
Main examples now are:
1. Tags to add to a receiving activity once payment is received.
2. Saving destination address for transactions sent before activity is created.

The PR also renames the previous use of "metadata tags" to just tags to fix previous confusion.


ovitrif (approved):
Great work, LGTM 🚀

PR #53, opened 2025-11-13
Add more fields to Pre Activity Metadata

ben-kaufman:
This PR adds fee rate, is transfer, and channel id to the PreActivityMetadata, and auto assigns them to an activity if found (like with the address).


jvsena42 (approved):
LGTM

PR #71, opened 2026-02-18
feat(trezor): add Trezor hardware wallet integration via trezor-connect-rs

coreyphillips:
This PR:

- Rewrites the trezor module (`src/modules/trezor/`) with `trezor-connect-rs`, enabling device discovery, connection, and Bitcoin operations over USB and Bluetooth

## Supported Operations

| Operation | Description |
|---|---|
| `trezor_scan` | Discover USB + BLE devices |
| `trezor_connect` / `trezor_disconnect` | Device connection lifecycle |
| `trezor_get_address` | Derive Bitcoin address (P2PKH, P2SH-P2WPKH, P2WPKH, P2TR) |
| `trezor_get_public_key` | Get xpub at BIP32 path |
| `trezor_sign_message` / `trezor_verify_message` | Message signing and verification |
| `trezor_sign_tx` | Full Bitcoin transaction signing with RBF support |

## Testing

- PR's will be created for both [iOS](https://github.com/synonymdev/bitkit-ios) & [Android](https://github.com/synonymdev/bitkit-android) versions of Bitkit containing a Trezor dev dashboard that will only be accessible in dev mode at the following location: Settings->Advanced->Trezor
- iOS
    - Only supports bluetooth.
    - No USB is available.
    - Trezor Safe 7 is the only device that supports bluetooth at the time of this writing.
- Android
    - Will support both bluetooth and USB.
    - Should support most if not all Trezor devices.


src/modules/trezor/types.rs
```diff
@@ -1,1006 +1,377 @@
...
+            passphrase_protection: f.passphrase_protection,
+            initialized: f.initialized,
+            needs_backup: f.needs_backup,
+        }
+    }
+}
+
+/// Script types for address derivation.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, uniffi::Enum)]
+pub enum TrezorScriptType {
```
ben-kaufman: Might want to add P2SH and P2WSH for future compatibility.

src/modules/trezor/types.rs
```diff
@@ -1,1006 +1,377 @@
...
-    pub amount: u64,
-    /// Script public key (binary hex)
-    pub script_pubkey: String,
+/// Parameters for getting an address from the device.
+#[derive(Debug, Clone, uniffi::Record)]
+pub struct TrezorGetAddressParams {
+    /// BIP32 path (e.g., "m/84'/0'/0'/0/0")
+    pub path: String,
+    /// Coin name (default: "Bitcoin")
+    pub coin: Option<String>,
```
ben-kaufman: We only ever want this to be Bitcoin (or BitcoinTestnet / Regtest if those are options), so I'd suggest to either remove or if testnet and/or regtest are different option make an enum for these.

ben-kaufman (approved):
utAck

PR #74, opened 2026-03-04
Add pubky module

ben-kaufman:
## Summary

- Adds a `pubky` module to bitkit-core (`src/modules/pubky/`) with auth flow orchestration and public file fetching, backed by the `pubky` SDK v0.6.0
- Auth flow (`start_pubky_auth` / `complete_pubky_auth` / `cancel_pubky_auth`) manages a global `PubkyAuthFlow` behind a `TokioMutex`, returning the `pubkyauth://` deep-link URL and awaiting Ring approval to produce a session secret for `paykit_import_session()`
- `fetch_pubky_file` fetches public resources from `pubky://` URIs through the SDK's PKARR/PubkyTLS transport, returning raw bytes — needed because `resolve_pubky_url` produces `https://_pubky.<z32>/…` hostnames that standard HTTP clients (URLSession, AsyncImage) cannot resolve

## New FFI exports

| Function | Type | Returns |
|---|---|---|
| `start_pubky_auth(caps)` | async | `pubkyauth://` URL |
| `complete_pubky_auth()` | async | session secret (`<pubkey_z32>:<cookie>`) |
| `cancel_pubky_auth()` | async | `()` |
| `resolve_pubky_url(uri)` | sync | `https://_pubky.<z32>/…` transport URL |
| `fetch_pubky_file(uri)` | async | `Vec<u8>` resource bytes |

## Test plan

- [x] `cargo check` passes with no new warnings
- [x] 6 unit tests pass (`cargo test --lib pubky`): resolve happy path (both URI forms), resolve error path, complete without active flow, cancel without active flow, fetch with malformed URI
- [x] End-to-end auth flow test with Pubky Ring (requires live relay + signing device)
- [x] Verify `fetch_pubky_file` loads a real profile image from a pubky homeserver

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -114,6 +114,24 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a23eb6b1614318a8071c9b2521f36b424b2c83db5eb3a0fead4a6c0809af6e61"
 
 [[package]]
+name = "argon2"
+version = "0.5.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3c3610892ee6e0cbce8ae2700349fcf8f98adb0dbfbee85aec3c9179d29cc072"
+dependencies = [
+ "base64ct",
+ "blake2",
+ "cpufeatures",
+ "password-hash",
+]
+
+[[package]]
+name = "arrayref"
+version = "0.3.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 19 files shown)


dzdidi:
So I am strongly considering to remove pubky-sdk re-export from paykit-lib. 
What do you think of going even further and also removing methods for getting profile and contacts? 
This will make bitkti as an app more responsible about how to do certain things (what is your evaluational opinion on this?) while will make paykit library less opinionated and  attached to the pubky.app which is a good thing in my opinion. 

https://github.com/pubky/paykit-rs/issues/25

PR #79, opened 2026-03-19
feat(onchain): add transaction history & detail

coreyphillips:
This PR:
- Adds `get_transaction_history` that queries an xpub's full transaction history and balance via Electrum/BDK.
- Adds `get_transaction_detail` that returns full data for a single transaction by txid.

- New types:
  - `HistoryTransaction`
  - `WalletBalance`
  - `TxDirection`
  - `TransactionHistoryResult`
  - `TransactionDetail`
  - `TxDetailInput`
  - `TxDetailOutput`


ovitrif (changes requested):
All previous review remarks addressed. One remaining item:

**Bindings need regeneration.** The `TransactionNotFound` error variant exists in Rust (`errors.rs:81`) but is missing from the Kotlin/Swift/Python bindings. The bindings were generated from the first commit but not re-generated after d45706d. Run the binding generation script and push updated bindings.

Once that's done, this is good to merge.

PR #85, opened 2026-04-01
fix: round up sub-satoshi invoice amounts to prevent underpayment

ben-kaufman:
## Summary
- Lightning invoices with millisatoshi precision (e.g. `18557538 msat` = `18557.538 sats`) were truncated by floor division to `18557 sats`, causing `sendUsingAmount` to send `18557000 msat` — less than the invoice requires — and LDK to reject the payment.
- Changed `amount_satoshis` from floor (`/ 1000`) to ceiling (`div_ceil(1000)`) so the sats value always covers the invoice's msat amount.
- Added unit tests covering the original failure, sub-satoshi remainders (above/below/equal `.5`), exact sat amounts, and boundary cases.

## Test plan
- [x] `cargo test scanner` — 19 tests pass
- [ ] E2E: pay invoices created with `lnd.addInvoice({ valueMsat })` using amounts `222538`, `222222`, `500500` msat
- [ ] Verify exact-sat invoices are unaffected

Closes synonymdev/bitkit-ios#511
Closes synonymdev/bitkit-android#877


ovitrif (approved):
utAck

PR #86, opened 2026-04-08
feat: support USB THP devices from trezor-connect-rs

coreyphillips:
This PR:
Resolves Trezor Safe 7 via USB error
- Query `transport.has_thp()` after acquire to correctly set `uses_thp` for any device that negotiated THP, not just Bluetooth.

New release for testing created [here](https://github.com/synonymdev/bitkit-core/releases/tag/v0.1.57).


ovitrif (approved):
utAck

ovitrif (approved):
The work here is on point; it leads to a better understanding on the issue by my 2 AIs, as they both agree this is a good step in the right direction.

Unfortunately, I still have an issue, and I can't really connect via USB yet, but the new error I'm hitting is now "earlier" in the flow, more self-explanatory and isolated much better.

I'm trying to pick up on this work and these insights and work on a fix, but it will be first in a fork of `trezor-connect-rs`.

PR #87, opened 2026-04-09
fix: pin trezor safe7 transport

ovitrif:
### Description

- pin trezor-connect-rs to commit 214d0c542ad50992811df3a83ab3e1be20ec4cad from ovitrif/trezor-connect-rs
- pull in the USB Safe 7 THP transport fix that reopens the device after the legacy V1 probe before starting THP
- rebuild the Android native libraries so the published Gradle package uses the patched transport implementation

### Preview

- n/a

### QA Notes

- ran cargo test modules::trezor
- ran cargo build
- ran cargo clippy
- ran ./build.sh android
- ./build.sh android completed successfully, including publishToMavenLocal
- full cargo test in this repo still hit unrelated blocktank test failures/hangs, so validation was focused on the Trezor path


ovitrif:
didn't fix it, I retested on Android via local changes integrating this.

PR #91, opened 2026-05-25
feat(trezor): return typed PassphraseResponse

coreyphillips:
Migrates Trezor passphrase handling from a `String` to the typed `PassphraseResponse` enum and adds per-connection wallet selection, letting apps choose which wallet a connection opens, including deferring passphrase entry to the Trezor screen.

### Description

Bumps `trezor-connect-rs` 0.2.8 → 0.3.2 and adopts its new typed passphrase
callback and session-passphrase support, replacing the old stringly-typed API.

**Why (typed passphrase):** the previous `on_passphrase_request(...) -> String`
encoding used `""` to mean *cancel* and any non-empty string to mean *hidden
wallet*. This left no way to express the **standard wallet** case (`Some("")`
on the device — an empty passphrase that selects the standard wallet rather
than a hidden one). A user selecting the standard wallet would send `""`, which
was interpreted as *cancel*. The new `Standard` variant closes that gap.

**Why (wallet selection):** on THP devices (Safe 5/7) the passphrase is bound
to the session at creation time, not via a mid-operation prompt. To support
that — and on-device passphrase entry — connections now declare up front which
wallet to open.

#### Passphrase callback (typed)
- `on_passphrase_request` now returns `PassphraseResponse` (`Cancel` /
  `Standard` / `Hidden { value }` / `OnDevice`) instead of `String`.
- Exposes the upstream `PassphraseResponse` to UniFFI bindings via
  `#[uniffi::remote(Enum)]` rather than forking the type (trezor-connect-rs
  intentionally has no uniffi dependency). The variant list is kept in sync
  with upstream and guarded by tests.
- Simplifies `UiCallbackAdapter::on_passphrase_request` to a pass-through (both
  traits now return the same upstream type). PIN keeps the legacy
  `String → Option` encoding (no upstream enum exists for it).
- Adds `TrezorError::PassphraseCancelled`, mapped from the upstream
  `TcDeviceError::PassphraseCancelled`.

#### Per-connection wallet selection
- Adds a `selection: WalletSelection` parameter to `trezor_connect`
  (`Standard` / `Hidden { passphrase }` / `OnDevice`). On THP devices the
  passphrase is bound to the session at creation, so it must be supplied on
  every connect — there is no separate "set passphrase" step and nothing is
  cached between calls; reconnect with a different `selection` to switch
  wallets. On desktop the passphrase is still supplied via the UI callback, so
  `selection` is unused on that path.
- Holds the host passphrase in `Zeroizing` so the buffer is wiped on drop
  (adds the `zeroize` dependency). `OnDevice` carries no host passphrase so the
  Trezor prompts on its own screen.
- Surfaces `TrezorFeatures.passphrase_entry_capable`
  (`Capability_PassphraseEntry`) so apps can tell whether on-device entry is
  available.

#### Other
- Maps the now-`#[non_exhaustive]` upstream error enums exhaustively, including
  `DeviceError::InvalidState` (wrong passphrase for a remembered wallet).
- Adds 7 adapter/passphrase tests covering PIN and passphrase mapping,
  including the critical Standard-must-not-become-Cancel case, `OnDevice`
  mapping, and `on_device` flag forwarding.
- Bumps version 0.1.60 → 0.1.62 (`Cargo.toml`, `Package.swift` + checksum,
  `gradle.properties`) and regenerates Swift / Kotlin / Python bindings.

> ⚠️ **Breaking FFI changes.** Two signatures change:
> 1. `TrezorUiCallback.onPassphraseRequest` must return a `PassphraseResponse`
>    instead of a `String`.
> 2. `trezorConnect` now takes a required `selection: WalletSelection`
>    argument.
>
> Both require a coordinated change in the bitkit iOS/Android apps.

### Preview

N/A — Rust FFI library change, no UI surface.

### QA Notes

Automated:
- `cargo test modules::trezor` — all pass, including the new
  `test_passphrase_adapter_*` / `test_pin_adapter_*` cases.
- `cargo clippy` — no new warnings introduced by this change.

Manual (on-device, against the updated app binding):
- **Standard wallet:** select standard / empty passphrase → wallet derives and
  the operation proceeds (previously this was treated as cancel).
- **Hidden wallet (host entry):** enter a passphrase on the host → hidden
  wallet derives correctly.
- **On-device entry:** select `OnDevice` / enter the passphrase on the Trezor
  itself → the device prompts and the connection opens the matching wallet.
- **Cancel:** dismiss the passphrase prompt → operation aborts with
  `PassphraseCancelled`.
- **Wallet switch:** reconnect with a different `selection` → connection opens
  the other wallet.


coreyphillips:
Release for testing [here](https://github.com/synonymdev/bitkit-core/releases/tag/v0.1.62).

coreyphillips:
Android PR available for testing [here](https://github.com/synonymdev/bitkit-android/pull/973).

ovitrif:
Retriggered gradle publish actions workflow to get the package up for bitkit-android CI, after this PR which fixes the issue that prevented it to publish in the first run was merged:
- #93

ovitrif (approved):
tAck via:
- synonymdev/bitkit-android#973

PR #94, opened 2026-05-29
feat: add onchain event listener

coreyphillips:
Adds an on-chain transaction watcher for extended public keys. A caller starts a watcher for an xpub and receives typed `WatcherEvent`s via an `EventListener` callback.

### Description
- New `EventListener` trait (`#[uniffi::export(with_foreign)]`) for receiving typed callbacks, implementable from Swift/Kotlin/Python.
- New FFI functions: `onchain_start_watcher`, `onchain_stop_watcher`, `onchain_stop_all_watchers`.
- New `WatcherParams` record: `watcher_id`, `extended_key`, `electrum_url`, optional `network` / `account_type` (auto-detected from the key prefix when omitted), and optional `gap_limit` (defaults to 20).
- New `WatcherEvent` enum: `TransactionsChanged { transactions, balance, tx_count, block_height, account_type }`, `Error { message }`, `Disconnected { message }`, `Reconnected`.
- Implementation (`src/modules/onchain/listener.rs`):
  - Uses Electrum `blockchain.scripthash.subscribe` for detection; a short-timeout `ping()` drains pushed notifications off the socket without a fixed poll interval.
  - Runs the wallet + Electrum client (both `!Send`) on one dedicated `std` thread per watcher rather than occupying a Tokio blocking-pool slot for the watcher's lifetime.
  - Init ordering is sync → subscribe → sync to close the sync/subscribe race window.
  - Self-healing connection: lost connections trigger a `Disconnected` event and an exponential-backoff reconnect (capped at 60s) that emits `Reconnected`; a failed resync is retried so a popped notification is never dropped.
  - Per-watcher generation guard so a tearing-down watcher can't evict a newer watcher reusing the same `watcher_id`.
  - Gap limit is extended automatically as new addresses are used.
- New `WatcherError` variant on `AccountInfoError`.
- Regenerated Android/iOS/Python bindings and bumped version to `0.1.65`.

### QA Notes
- `cargo check` and `cargo test` pass.
- Start a watcher against a known xpub on an Electrum server and confirm:
  - An initial `TransactionsChanged` event arrives with the current balance/history.
  - Sending a new tx to a derived address fires a `TransactionsChanged` update within seconds.
  - Killing/restoring the Electrum connection produces `Disconnected` then `Reconnected` events, and state resyncs afterward.
  - `onchain_stop_watcher` / `onchain_stop_all_watchers` tear the watcher(s) down cleanly.
  - Starting a second watcher with an already-active `watcher_id` returns a `WatcherError`.


coreyphillips:
- Release created as 0.1.65 [here](https://github.com/synonymdev/bitkit-core/releases/tag/v0.1.65).
- Android PR for testing is available [here](https://github.com/synonymdev/bitkit-android/pull/976).

src/modules/onchain/listener.rs
```diff
@@ -0,0 +1,724 @@
...
+            return;
+        }
+
+        // ping() keeps the connection alive and blocks on the socket, reading
+        // and dispatching any push notifications that arrive before/with the
+        // ping response. With the SOCKET_TIMEOUT_SECS read timeout it either:
+        // - returns Ok(()) quickly if there's socket activity (notifications!),
+        // - returns Err(timeout) after the timeout when idle, or
+        // - returns Err(other) on a real connection failure.
+        match sub_client.ping() {
```
ben-kaufman, line 683: `ping()` returns as soon as the Electrum server replies, so an idle watcher loops immediately and sends another `server.ping`. That can burn CPU and get the server to rate limit or disconnect us. Could we add an explicit idle wait when nothing was popped, or switch this to a real blocking notification read?

src/modules/onchain/listener.rs
```diff
@@ -0,0 +1,724 @@
...
+/// rebuilds it — making the loop self-healing without a dedicated reconnect for
+/// the sync channel.
+fn ensure_synced(
+    wallet: &Wallet<MemoryDatabase>,
+    blockchain: &mut Option<ElectrumBlockchain>,
+    electrum_url: &str,
+) -> Result<(), AccountInfoError> {
+    if blockchain.is_none() {
+        let client = connect_electrum(electrum_url)?;
+        *blockchain = Some(ElectrumBlockchain::from(client));
```
ben-kaufman: `gap_limit` is used when subscribing scripts, but the wallet refresh still uses `ElectrumBlockchain::from(client)`, which syncs with BDK's default stop gap of 20. If a caller sets `gap_limit` above 20, a farther address can trigger a notification but still be missing from the refreshed wallet state. Could we pass the watcher gap into the Electrum blockchain sync path and add a focused test for this?

ben-kaufman (commented):
Found two watcher issues in the review.

src/modules/onchain/listener.rs
```diff
@@ -0,0 +1,724 @@
...
+    subscribe_scripts_reporting(&sub_client, &subscribed_scripts, &listener, &watcher_id);
+
+    // Second sync: brackets the subscribe to close the sync/subscribe race.
+    if let Err(e) = ensure_synced(&wallet, &mut blockchain, &params.electrum_url) {
+        let _ = init_tx.send(Err(e));
+        return;
+    }
+
+    // Init succeeded.
+    let _ = init_tx.send(Ok(()));
```
ben-kaufman, line 632: nit: `stop_watcher` can remove the handle and send shutdown while `start_watcher` is still waiting on init, but startup does not check `shutdown_rx` until after it sends `Ok(())` and the initial event. That means a watcher can report a successful start and call the listener after it was already stopped. Could we check shutdown before reporting init success and before sending the initial event?

ben-kaufman (commented):
Found one small lifecycle edge in the second pass.

src/modules/onchain/listener.rs
```diff
@@ -0,0 +1,724 @@
...
+}
+
+/// Subscribe to all scripts, best-effort (used on reconnect).
+///
+/// Subscribe failures are ignored here because the resync that always follows
+/// a reconnect re-reads full wallet state regardless.
+fn subscribe_scripts(client: &electrum_client::Client, scripts: &HashSet<ScriptBuf>) {
+    for script in scripts {
+        let _ = client.script_subscribe(script);
+    }
```
ben-kaufman: nit/perf: `electrum-client` has `batch_script_subscribe`, so this loop and the startup loop in `subscribe_scripts_reporting` could use one batched call instead of one RPC per script. That would make watcher startup and reconnect cheaper when the gap is large. Could we use the batch API here, or is there a reason we need per-script subscription calls?

ben-kaufman (commented):
One small perf note from the BDK/electrum-client pass.

src/modules/onchain/listener.rs
```diff
@@ -0,0 +1,724 @@
...
+// ============================================================================
+// Watcher configuration
+// ============================================================================
+
+/// Parameters for starting an xpub transaction watcher.
+#[derive(Debug, Clone, uniffi::Record)]
+pub struct WatcherParams {
+    /// Caller-supplied identifier for this watcher.
+    pub watcher_id: String,
+    /// Extended public key (xpub/ypub/zpub/tpub/upub/vpub).
```
ben-kaufman, line 48: design question: BDK is descriptor-first, but the watcher API takes an xpub and reconstructs descriptors from local assumptions about script type and the 0/1 branches. Would it be cleaner for the watcher setup to accept external and internal descriptors directly, with a helper that converts the current xpub params into descriptors for convenience? That would make this more future proof if we add multisig, miniscript, or other descriptor-based wallets later.

ben-kaufman (commented):
One design question about keeping the watcher descriptor-first.

src/modules/onchain/listener.rs
coreyphillips: Addressed in [this](https://github.com/synonymdev/bitkit-core/pull/94/changes/caf20e319bd82733ec31002b299c2c654e0a1c2f) commit.

src/modules/onchain/listener.rs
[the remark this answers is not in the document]
coreyphillips: Addressed in [this](https://github.com/synonymdev/bitkit-core/pull/94/changes/a79e7133cd25ed3aac76fef9301b5558119660ab) commit.

coreyphillips: Addressed in [this](https://github.com/synonymdev/bitkit-core/pull/94/changes/a79e7133cd25ed3aac76fef9301b5558119660ab) commit.

src/modules/onchain/listener.rs
coreyphillips: Addressed in [this](https://github.com/synonymdev/bitkit-core/pull/94/changes/6f8129dcf27e9966df88730eff5116c6bdb9be68) commit.

src/modules/onchain/listener.rs
coreyphillips: Addressed in [this](https://github.com/synonymdev/bitkit-core/pull/94/changes/e341210a1bd07a28c901e0f9d7be936ffd0c6950) commit.

src/modules/onchain/listener.rs
coreyphillips: I agree with that direction, but we might want to make the change in a separate PR so we can apply it module-wide.

src/modules/onchain/listener.rs
```diff
@@ -0,0 +1,812 @@
...
+            return;
+        }
+    };
+
+    // Persistent blockchain reused for every resync (see `ensure_synced`).
+    let mut blockchain: Option<ElectrumBlockchain> = None;
+
+    // First sync: establishes address usage so `derive_scripts` covers the
+    // full used range.
+    if let Err(e) = ensure_synced(&wallet, &mut blockchain, &params.electrum_url, sync_stop_gap) {
```
ben-kaufman: nit: `cargo fmt --check` fails on these new `ensure_synced(...)` calls. Could we run `cargo fmt` so the Rust formatting check passes?

ben-kaufman (commented):
One small formatting note

src/modules/onchain/listener.rs
```diff
@@ -0,0 +1,812 @@
...
+                            break;
+                        }
+                        None => return,
+                    }
+                }
+            }
+        }
+
+        if needs_resync {
+            if let Err(e) = ensure_synced(&wallet, &mut blockchain, &params.electrum_url, sync_stop_gap) {
```
ben-kaufman: If `ensure_synced(...)` fails while `needs_resync` is true, this `continue` skips the idle wait at the bottom of the loop and retries on the next iteration with `needs_resync` still set. If the sync side keeps failing quickly while the subscription client stays connected, this can repeatedly ping and emit `Error` events without backoff. Could we add a small retry backoff or reuse the idle wait before continuing?

ben-kaufman (commented):
One retry pacing note from the follow-up review.

src/modules/onchain/listener.rs
```diff
@@ -0,0 +1,812 @@
...
+fn subscribe_scripts_reporting(
+    client: &electrum_client::Client,
+    scripts: &HashSet<ScriptBuf>,
+    listener: &Arc<dyn EventListener>,
+    watcher_id: &str,
+) {
+    if scripts.is_empty() {
+        return;
+    }
+    if let Err(e) = client.batch_script_subscribe(scripts.iter().map(|s| s.as_script())) {
```
ben-kaufman: If the initial `batch_script_subscribe` fails, we emit an `Error` event but still continue startup and can return `Ok(())`. In that state the watcher may only refresh on block header notifications, so near-instant script updates are not actually active. Should the initial subscription failure make startup fail instead, so `Ok(())` means the watcher is fully subscribed?

ben-kaufman (commented):
One startup semantics question from the architecture pass.

src/modules/onchain/listener.rs
```diff
@@ -0,0 +1,812 @@
...
+                listener.on_event(
+                    watcher_id.clone(),
+                    WatcherEvent::Disconnected {
+                        message: format!("Connection lost: {}", e),
+                    },
+                );
+                match reconnect_loop(&params.electrum_url, &shutdown_rx, &subscribed_scripts) {
+                    Some((new_client, new_tip)) => {
+                        sub_client = new_client;
+                        tip_height = tip_height.max(new_tip);
```
ben-kaufman: nit: On reconnect we keep the previous height if the new Electrum connection reports a lower tip. Since `tip_height` is used to compute confirmations, reconnecting to a lagging server, load-balanced backend, or reset regtest node can overstate confirmations. Could we use the new connection's tip height instead, or otherwise make the emitted `block_height` match the chain view we just reconnected to?

ben-kaufman (commented):
One tip-height edge case from the architecture pass.

src/modules/onchain/listener.rs
coreyphillips: Addressed in [this](https://github.com/synonymdev/bitkit-core/pull/94/changes/76e23d149faa03a7bcba616ef9b403520dbe96b7) commit.

src/modules/onchain/listener.rs
ben-kaufman: Ok sure, let's do it separately

src/modules/onchain/listener.rs
coreyphillips: Addressed in [this](https://github.com/synonymdev/bitkit-core/pull/94/changes/cd24fbcf21059ccdd5b853ab89746a640c586a2f) commit.

src/modules/onchain/listener.rs
coreyphillips: Addressed in [this](https://github.com/synonymdev/bitkit-core/pull/94/changes/132dcb02ad270285ab3006454b245765f08dd124) commit.

src/modules/onchain/listener.rs
coreyphillips: Addressed in [this](https://github.com/synonymdev/bitkit-core/pull/94/changes/692eb97fccda6a7ff1e8e113bfc353ad8d906a02) commit.

src/modules/onchain/listener.rs
```diff
@@ -0,0 +1,847 @@
...
+/// Subscribe to all scripts, best-effort (used on reconnect).
+///
+/// One batched RPC instead of one call per script, so reconnect stays cheap even
+/// with a large gap. Failures are ignored here because the resync that always
+/// follows a reconnect re-reads full wallet state regardless.
+fn subscribe_scripts(client: &electrum_client::Client, scripts: &HashSet<ScriptBuf>) {
+    if scripts.is_empty() {
+        return;
+    }
+    let _ = client.batch_script_subscribe(scripts.iter().map(|s| s.as_script()));
```
ben-kaufman: nit: Could we surface subscription failures here too, or keep retrying reconnect until the script subscriptions and header subscription both succeed? `electrum-client` registers the local notification queues before the RPC finishes, so if `batch_script_subscribe` fails and we ignore it, `script_pop` can look normal even though the server-side script subscriptions may not be active. Same idea for the gap-extension `script_subscribe` calls below: maybe only insert the script after subscribe succeeds.

ben-kaufman (commented):
One small subscription nit.

src/modules/onchain/listener.rs
coreyphillips: Addressed in [this](https://github.com/synonymdev/bitkit-core/pull/94/changes/964c4d012e967eac80f2213f91f8accd6c9ae5f8) commit.

ben-kaufman (approved):
LGTM, would be good to update to more descriptor focus architecture as a follow up.

coreyphillips:
I'll bump the version and rebuild.

- New v0.1.66 release [here](https://github.com/synonymdev/bitkit-core/releases/tag/v0.1.66).

PR #103, opened 2026-06-19
feat: add wallet-scoped activity storage

### LINKED ISSUE #102: feat: add wallet-scoped hardware wallet activities
## Context

Bitkit Android is adding Trezor hardware-wallet UI in synonymdev/bitkit-android#998. The Android subtask synonymdev/bitkit-android#1029 needs Core support so hardware-wallet watcher transactions can reuse the normal activity stack: stable activity detail loading, inputs/outputs, tags, unified activity lists, and cleanup when a paired hardware wallet is removed.

Today Android has to keep hardware-wallet watcher activity outside the Core activity DB, manually merge it with Core-backed activity lists, sort/filter/paginate in app code, and fall back on reduced detail behavior. This should move into Core so Android and iOS can share the same activity behavior.

Keep this simple: `walletId` can be an opaque persisted identifier for a wallet, including a randomly generated id supplied by the app. It does not need to be deterministically derived from a descriptor or Trezor public material in this task. Apps own hardware-wallet metadata, backup/restore, and reattach behavior; Core only needs to preserve and operate on the caller-provided `walletId`.

## Scope

Add wallet-scoped activity support to Bitkit Core.

### Activity storage

Add a `walletId` field to activity records so Core can distinguish activity belonging to different wallets.

Requirements:

- Existing Bitkit wallet activity should keep working with minimal migration, e.g. `walletId = null`, a default wallet id, or another simple backward-compatible representation.
- Hardware-wallet activity should be stored with the caller-provided `walletId`.
- Activities from different wallets must not collide when they share the same `txid`.
- Returned activity records should include enough wallet/source metadata for clients to render hardware-wallet styling and navigate to the correct detail flow.

### Activity APIs

Expose binding APIs that support both per-wallet and unified activity usage:

- Upsert hardware-wallet watcher activity with a `walletId`.
- Query activities for one `walletId`, e.g. a hardware-wallet detail screen.
- Query the normal all-activities feed with Bitkit wallet activity and hardware-wallet activity in the same result set.
- Sort the unified activity feed after combining wallet scopes so `limit`/pagination returns the true latest activity across Bitkit and hardware wallets.
- Preserve `walletId` on returned activity records.
- Delete all activity/details/tag metadata for one removed hardware-wallet `walletId`.

Existing-style activity filters should continue to work on the unified list where applicable: filter, tx type, tags, search, date range, limit, and sort direction.

### Details and tags

Hardware-wallet activity should reuse normal activity detail and tag behavior.

Requirements:

- Activity Detail can load transaction inputs and outputs for hardware-wallet transactions using the configured Electrum/electrs backend.
- Tags can be added/removed on hardware-wallet activities through the same user-facing behavior as normal activities.
- Tags and other mutable metadata must not leak between wallets when two wallet-scoped activities share a `txid`.
- Cleanup for a removed `walletId` should delete scoped activity metadata without deleting data still needed by another wallet's activity.

### Watcher ingestion

Support converting hardware-wallet watcher history into Core on-chain activities under a supplied `walletId`.

Requirements:

- Multiple account/address-type watchers under the same `walletId` should merge into one visible activity per transaction where appropriate.
- A transaction seen by both P2WPKH and Taproot watchers for the same hardware wallet must not create duplicate visible activity rows.
- A transaction that also belongs to Bitkit's main wallet must not overwrite or corrupt the main wallet activity/details/tags.

## Acceptance criteria

- Core can persist and return activities with `walletId` metadata.
- Clients can upsert hardware-wallet activities with an opaque `walletId`.
- Existing Bitkit wallet activity behavior remains backward compatible.
- The all-activities query can return Bitkit wallet activity and hardware-wallet activity in one sorted list.
- Unified list filters, sorting, limits, and pagination behave consistently after combining wallet scopes.
- Per-wallet queries return only activity for the requested `walletId`.
- Same-`txid` activity can exist for different wallets without collisions or metadata leaks.
- Hardware-wallet activity details include transaction inputs and outputs when available from the configured Electrum/electrs backend.
- Tags work for hardware-wallet activity and remain scoped correctly.
- Removing a hardware wallet deletes Core activity/details/tag metadata scoped to that `walletId` without affecting other wallets.
- Tests cover opaque `walletId` storage, unified list sorting/pagination, per-wallet queries, same-`txid` collision isolation, tag behavior, detail cleanup, and FFI/binding exposure.

ben-kaufman:
## Summary

- Add `wallet_id` to activity, activity tag, and transaction detail data so multiple wallets can safely store overlapping activity IDs or transaction IDs.
- Rebuild activity-related primary keys and migrations around wallet scope, defaulting pre-existing rows to the default Bitkit wallet ID.
- Update activity/tag/transaction-detail APIs to accept wallet IDs where scope matters, while keeping global queries global where intended.
- Harden tests for wallet-scoped collisions, migrations, tag backup/restore, transaction details, and preserving tags/seen state through upserts.
- Update activity docs/examples and regenerate iOS, Android, and Python bindings with the minor version bump to `0.3.0`.

Closes #102.

## Validation

- `cargo test modules::activity` passed: 155 tests
- `./build.sh -r --minor all` performed the version bump; after the initial sandbox DNS failure, `./build.sh all` completed successfully
- Android Maven local publish check passed during the build
- `git diff --check` passed


ovitrif:
Seems ok architecturally;

Though if noticed that it still looks up for pre-activity metadata by address or payment id only, then writes the tags into the wallet of the activity being inserted.

I reproduced this with Codex using the Python binding: metadata added for an address was attached to a `hardware-wallet-1` activity with the same address.

Could we add `wallet_id` to `PreActivityMetadata` and `pre_activity_metadata`, default legacy rows to `bitkit`, and scope lookup/delete by wallet too?

ovitrif:
The top-level README still shows the old activity signatures without `wallet_id`, but this PR changes the exported activity API. Could we update this section too and mention `get_default_wallet_id()` so consumers following the root docs get the scoped API?

coreyphillips (approved):

ovitrif (commented):
Re-reviewed the latest push. The original `PreActivityMetadata` wallet-scope issue and the root README activity API docs issue are mostly addressed. I found a few follow-up points:

1. `src/modules/activity/implementation.rs:897`

Could we make the bulk upsert path apply pending pre-activity metadata too? `insert_activity` transfers and deletes metadata, but `upsert_onchain_activities`, `upsert_lightning_activities`, and `upsert_activities` just write rows. I reproduced this through the Python binding: metadata tags stay pending and the inserted activity has no tags after `upsert_onchain_activities`. Android watcher sync is likely to use this bulk path, so hardware activities would miss the same tags and metadata support this PR is adding.

2. `src/modules/activity/implementation.rs:2546`

Could we include `is_receive = 1` in the address lookup used for received onchain transfers? Right now a metadata row for the same address with `is_receive = false` is selected for a received activity, then cleanup leaves it behind because it only deletes `is_receive = 1`. I reproduced this through the Python binding: the received activity got the sent-only tag and the pending row remained.

3. `src/lib.rs:1685`

Should `is_address_used` take a `wallet_id` now too? It still counts any onchain activity matching `address` across all wallets. I reproduced a hardware-only activity making `is_address_used(address)` return `true` even though the default wallet had no matching activity. That can make the Bitkit wallet treat a hardware-wallet address as already used.

4. `src/modules/activity/README.md:286`

Could we re-check these examples against the generated bindings? The Kotlin enum payloads are under `v1`, so `foundActivity.txId` and `foundActivity.preimage` will not compile, and there is an extra `}` after the catch block. The Python sample also omits required `OnchainActivity` fields like `contact`, `created_at`, `updated_at`, and `seen_at`, and some `get_activities` calls omit nullable parameters that the generated Python API still requires.

Local checks:

- `cargo fmt --check`: passed
- `cargo test modules::activity`: passed, 159 tests

Leaving this as a neutral review.

ben-kaufman:
Thanks Ovi, pushed a follow-up in 8cde6b8 covering the review points:

1. Bulk upserts now apply pending pre-activity metadata after the DB transaction commits for both onchain and lightning paths, so watcher-style restores/syncs attach tags and clear pending metadata the same way inserts do.
2. Received onchain metadata lookup is now constrained to `is_receive = 1`. I also tightened add/restore replacement semantics so receive metadata only replaces previous receive metadata for the same wallet/address, while sent metadata using the same address can coexist.
3. I left `is_address_used(address)` global intentionally. Since an address is globally unique, if any wallet has used it then it is used; a wallet-scoped variant feels like a separate helper/API decision rather than part of this PR.
4. Rechecked and updated the README examples/docs: Kotlin uses enum payloads via `v1`, Python includes the required activity fields and nullable `get_activities` params, and the root/activity API docs include the wallet-scoped signatures.

Added regression coverage for the update/bulk metadata transfer and receive-address replacement cases. Latest local checks: `cargo fmt --check`, `git diff --check`, and `cargo test modules::activity` all pass (`164 passed`).

bindings/python/bitkitcore/__init__.py
```diff
@@ -0,0 +1 @@
+from .bitkitcore import *
```
ovitrif, line 1: Could we regenerate the checked-in native bindings/artifacts after the latest Rust fixes? The Rust source now applies pending metadata in bulk upserts and scopes received-address lookup by `is_receive = 1`, but the committed Python native library still reproduces the previous behavior when loaded from `bindings/python`: `upsert_onchain_activities` leaves the pending metadata behind, and a sent-only address metadata row still attaches to a received activity. Since the latest two fix commits did not update the checked-in native artifacts, Android/iOS consumers may still get stale behavior too until the bindings and native libs are rebuilt.

src/modules/activity/implementation.rs
```diff
@@ -1190,6 +1684,8 @@ impl ActivityDB {
             error_details: format!("Failed to commit transaction: {}", e),
         })?;
 
+        self.apply_pre_activity_metadata_for_onchain(activity);
```
ovitrif: Could we make the post-update metadata restore target the row selected by the `activity_id` argument, or reject mismatched IDs? `update_activity(activity_id, activity)` updates the existing row by `activity_id`, but this new restore call passes only `activity`, and the helper uses `activity.id` as the tag/metadata target. If those values differ, pending metadata can be attached to the wrong/nonexistent activity ID or deleted without reaching the row that was actually updated.

ovitrif (commented):
Re-reviewed the latest push. The stale bulk-upsert, received-address lookup, and activity README example points are fixed in the Rust source, so I’m only leaving the two current follow-ups I could still verify.

Local checks:

- `cargo fmt --check`: passed
- `git diff --check origin/master...HEAD`: passed
- `cargo test modules::activity`: passed, 164 tests
- Python binding smoke test: import works, but the committed native artifact still reproduces the old metadata behavior noted below

Leaving this as a neutral review.

src/modules/activity/implementation.rs
ben-kaufman: Fixed in aa6cf68. The post-update metadata restore now takes an explicit target `activity_id`: update methods pass the `activity_id` argument, while bulk paths pass each activity’s own ID. I added onchain and lightning regression tests with mismatched payload IDs to verify tags land on the row selected by the update call and pending metadata is cleared.

bindings/python/bitkitcore/__init__.py
ben-kaufman: Fixed in aa6cf68. I ran `./build.sh all` and committed refreshed native artifacts: Android `.so` files, iOS XCFramework/static libs/zip plus the Package.swift checksum, and Python `libbitkitcore.dylib`. I also smoke-tested the checked-in `bindings/python` package against the stale-artifact cases: bulk upsert now transfers/clears pending metadata, and sent-only address metadata is not attached to a received activity.

ovitrif (approved):
LGTM, thanks for addressing the follow-ups.

PR #104, opened 2026-06-19
feat: expose trezor lock state

ovitrif:
Related: synonymdev/bitkit-android#1030, synonymdev/bitkit-android#998

### Description

- Expose Trezor lock state via `TrezorFeatures.unlocked` so mobile clients can derive `pinProtection == true && unlocked == false` for locked devices.
- Add `trezor_refresh_features()` as an explicit one-shot feature refresh that updates cached features without adding polling.
- Surface busy transport state as `TrezorError::DeviceBusy` so clients can back off while the device is busy or awaiting user action.
- Add `TrezorTransportErrorCode.DeviceBusy` to native callback results so mobile transports can report busy open/read/write/message calls without relying on generic error strings.
- Release: [v0.3.2](https://github.com/synonymdev/bitkit-core/releases/tag/v0.3.2)

### QA Notes

- `cargo fmt --check`
- `cargo test modules::trezor`
- `cargo clippy`
- `./build.sh -r --patch all`
- `./build.sh all`
- Published release: [v0.3.2](https://github.com/synonymdev/bitkit-core/releases/tag/v0.3.2)
- Release assets uploaded: `BitkitCore.xcframework.zip`, `native-debug-symbols.zip`


src/modules/trezor/errors.rs
```diff
@@ -103,9 +107,7 @@ impl From<trezor_connect_rs::TrezorError> for TrezorError {
             TE::Transport(transport_err) => match transport_err {
                 TcTransportError::DeviceNotFound => TrezorError::DeviceNotFound,
                 TcTransportError::DeviceDisconnected => TrezorError::DeviceDisconnected,
-                TcTransportError::DeviceBusy => TrezorError::ConnectionError {
-                    error_details: "Device is busy".to_string(),
-                },
+                TcTransportError::DeviceBusy => TrezorError::DeviceBusy,
```
ben-kaufman, line 141: `DeviceBusy` only seems to be surfaced when upstream returns `TransportError::DeviceBusy`. On the mobile callback path, a failed native `open_device` is converted to `UnableToOpen`, and failed message calls become `DataTransfer`, so a busy native transport result still looks generic to clients. Could we map the callback busy case to `TrezorError::DeviceBusy` too, or add a structured busy code to the callback result?

ovitrif: Added structured `TrezorTransportErrorCode.DeviceBusy` to the native callback result records and map that code through the callback adapter before upstream converts mobile callback failures to `UnableToOpen`/`DataTransfer`. Those callback busy failures now convert back to `TrezorError::DeviceBusy`, with tests for both open and data-transfer paths.

Resolved in d95a061

ben-kaufman (approved):
Reaudited latest head d95a061. The callback DeviceBusy issue is fixed and checks pass.

Issue #107, opened 2026-06-23
Updates to facilitate Trezor integration in mobile


ovitrif:
> While reviewing the watch-only hardware-wallet layer on this branch, a few things stood out that would be better owned by core than reimplemented per-platform.
> 
> Note: I diffed against the local `bitkit-core` checkout, which is at **0.1.40** and predates the watcher / wallet-scoped-activity surface this PR consumes from **0.3.2** — so these are against the consumed 0.3.2 binary, not the source. Flagging in case the source needs a bump too.

## Subtasks
- [x] 1. **Core should own the `wallet_id` derivation**
      The per-device id is improvised on the Swift side (`HwWalletId.derive` = `"trezor:" + SHA256(sorted xpubs)`) and the PR already flags it as interim pending agreement with Android. If iOS and Android derive it independently, the same physical device yields different ids and activity won't reconcile cross-platform. Suggest a core `deriveWalletId(xpubs:)` so both platforms are guaranteed identical by construction.
- [x] 2. **Have the watcher emit `Activity` instead of `HistoryTransaction`**
      `HwWalletManager.onchainActivity()` hand-reconstructs an `Activity` from `HistoryTransaction` — guessing direction from `received > sent`, netting the fee, and hardcoding `feeRate: 1` / `address: ""`. That's core's job, and the placeholder fee-rate will render wrong. Emitting `Activity` (or exposing a `watchOnlyActivityFromHistory(walletId:txs:)` mapper) removes a per-platform mapping that iOS and Android will otherwise drift on.
- [x] 3. **`gapLimit: 20` should be a core default**
      instead of being duplicated in `HwWalletManager` and the watcher params on each platform.
- [ ] 4. **Core should own supported hardware-wallet metadata and capabilities**
      Android currently carries supported-device display metadata in app code while Trezor is only one concrete implementation of the generic hardware-wallet layer. Expose the supported vendors/models/capabilities from core so Android and iOS can render supported devices generically instead of duplicating Trezor-specific supported-HW info per platform. Follow-up from synonymdev/bitkit-android#1039.

_Originally posted by @jvsena42 in https://github.com/synonymdev/bitkit-ios/issues/605#issuecomment-4778838344_

ovitrif:
@coreyphillips I added subtask `4`, it's not as critical as the others but also can't be delayed too much 🙏🏻 

Should be an easy fix though, basically just modeling the type(s) of supported HWWs vendors, only Trezor for now:

[...]

PR #108, opened 2026-06-23
feat: core derive_wallet_id for hardware watch-only wallets

coreyphillips:
## Summary

Subtask 1 of #107. Core now owns the `wallet_id` derivation for hardware (watch-only) wallets so iOS and Android produce the **same id for the same physical device by construction**, instead of each platform reimplementing it. The Swift side currently improvises `HwWalletId.derive` = `"trezor:" + SHA256(sorted xpubs)` and flagged it as interim pending agreement with Android — if the two platforms derive independently they drift and activity won't reconcile cross-platform.

## What changed

- New `#[uniffi::export] pub fn derive_wallet_id(device_type: String, xpubs: Vec<String>) -> String`.
- Device-agnostic (`device_type` keeps `trezor` / `ledger` / … distinct), so it generalises beyond Trezor.

### Canonical derivation
1. sort `xpubs` lexicographically (input order doesn't matter),
2. join with a single `\n`,
3. SHA256 the UTF-8 bytes, hex-encode lowercase,
4. return `"{device_type}:{hash}"` (e.g. `trezor:ab12…`).

Trezor watch-only is new, so core defines the canonical form; platforms call this instead of rolling their own.

## Tests

`cargo test derive_wallet_id` — order-independence, device_type distinctness, a fixed known-vector, and the empty-xpubs case.


ben-kaufman (commented):
Two things I think we should tighten before merging:

1. `derive_wallet_id` should reject an empty `xpubs` list instead of returning `trezor:<sha256 empty string>`. Otherwise any caller that reaches this with no account xpubs gets the same wallet id for that device type, which can collapse separate setup failures into the same activity scope. Could we make this return an error and update the empty-input test?

2. The bindings look stale or partially regenerated. The Python wrapper references `uniffi_bitkitcore_fn_func_derive_wallet_id`, but the committed dylib does not export it, so `python3 -c "import bitkitcore"` fails with `symbol not found`. The Swift wrapper also calls the new symbol while the iOS headers and XCFramework do not include it, and Android has no `deriveWalletId` wrapper or JNI declaration. Could we regenerate and commit all platform bindings and artifacts from this Rust head?

coreyphillips:
Thanks @ben-kaufman, both addressed in 80dadad.

1. `derive_wallet_id` now returns `Result<String, ActivityError>` and rejects an empty `xpubs` list (also blank entries and a blank `device_type`), so an empty set can no longer collapse separate devices into one activity scope. The empty-input test was updated to assert the error.

2. Regenerated all platform bindings and native artifacts from this Rust head via `build.sh all`: Swift + `bitkitcoreFFI.h` + the XCFramework, Kotlin + the `jniLibs` `.so` files, and the Python wrapper + `libbitkitcore.dylib`. The new export is now present and callable in all three, so `import bitkitcore` no longer fails with a missing symbol.

ben-kaufman (approved):
Looks good, thanks for fixing the empty `xpubs` case and regenerating the bindings.

One release note before publishing: since the regenerated Swift zip checksum changed while `Package.swift` still points at `v0.3.3`, we may need to bump the bindings/package version so SwiftPM does not fetch the existing `v0.3.3` asset with the old checksum.

ovitrif:
also LGTM 👍🏻 🎉

PR #109, opened 2026-06-23
feat: watcher emits persistence-ready watch-only activities

coreyphillips:
## Summary

Subtask 2 of #107, reworked per @ovitrif's review. The onchain xpub watcher now emits **persistence-ready Core activity data** that Bitkit stores through the existing activity APIs — rather than the app hand-reconstructing activities per-platform (and drifting).

## What it does

**Watcher builds the payload (where the data still exists).** `WatcherEvent::TransactionsChanged` now carries:
- `activities: Vec<Activity>`
- `transaction_details: Vec<TransactionDetails>`

…instead of `Vec<HistoryTransaction>`. They're built inside the watcher using the BDK `Wallet`, raw txs (`list_transactions(true)`), confirmation time, fee/vsize, and `is_mine` ownership — the decode path is shared with `get_transaction_detail` via the extracted `map_bdk_tx_to_detail`.

**Wallet boundary = the watched address type.** `WatcherParams` gains `wallet_id`. One activity per tx, scoped to that wallet; the same txid under two address types becomes two wallet-scoped activities (no cross-address-type merge).

**Persistence-compatible semantics, from the watched wallet's perspective:**
| case | value | fee | address |
|------|-------|-----|---------|
| Received | `received` | `0` (sender paid it) | owned output |
| Sent | `sent - received - fee` | `fee` | destination (non-owned) output |
| SelfTransfer | `0` | `fee` | owned output |

Self-transfer is `value = 0, fee = fee` so Bitkit's `value + fee` total isn't doubled. Unconfirmed txs get a positive (`now`) timestamp so rows are DB-valid and sort to the top while pending.

**Upsert hardening** so watcher refreshes don't clobber user/first-seen state: `contact` is preserved when the incoming value is `None` (COALESCE), and an unconfirmed tx keeps its existing timestamp until it confirms. (`seen_at` and tags were already preserved.)

**Also:** `HistoryTransaction` keeps the new `fee_rate` field for `get_transaction_history` consumers; both that path and the watcher now use `list_transactions(true)`. The earlier standalone `watch_only_activity_from_history` mapper is removed.

## Tests

`cargo test --lib modules::onchain::` (55) and `modules::activity::` (167) pass. New coverage:
- `watch_only_activity_from_detail`: received-drops-sender-fee, sent value/fee split, self-transfer zero-value, unconfirmed `now` timestamp, matching `TransactionDetails`.
- `test_upsert_preserves_contact_and_unconfirmed_timestamp`.

Swift + Python bindings regenerated (iOS xcframework / Android Kotlin / dylib rebuilt at release via build.sh, per repo convention).


coreyphillips:
## Update: aligned with bitkit-ios / bitkit-android conventions

Checked how both apps key and build onchain activities to settle the open `id` question:

- **`id = tx_id`** — confirmed. iOS `CoreService.createSentOnchainActivityFromSendResult` uses `id: txid`, and Android `HwWalletRepo.toOnchainActivity` uses `id = first.txid`. The activity table's PK is `(wallet_id, id)`, so keying on the bare txid is collision-safe across wallets. No change needed.
- **`address = ""`** — matches Android's watch-only mapping exactly.

While there I found the apps do one more thing the 1:1 mapper didn't: **they merge `HistoryTransaction` rows by txid** (Android `toMergedActivities`). A hardware device has multiple accounts/script-types watched separately, so the same txid can surface once per account (e.g. an internal send→receive). With `id = tx_id`, two un-merged rows would collapse under `upsert` to last-wins and lose the combined picture. The mapper now:

- groups rows by txid, sums `received`/`sent`, and classifies the whole tx via core's `classify_tx` (the canonical direction/amount logic);
- keeps the real `fee`/`fee_rate` from the account that paid (received-only rows carry none);
- preserves first-seen txid order; produces exactly one activity per tx.

So the mapper is now a faithful drop-in for the per-platform logic rather than just a field copy. Tests cover received / sent / self-transfer / same-txid merge / ordering / unconfirmed defaults / deterministic id.

src/lib.rs
```diff
@@ -470,6 +470,58 @@ pub fn get_default_wallet_id() -> String {
...
+                wallet_id: wallet_id.clone(),
+                id: tx.txid.clone(),
+                tx_type,
+                tx_id: tx.txid,
+                value: tx.amount,
+                fee: tx.fee.unwrap_or(0),
+                fee_rate: tx.fee_rate.map(|r| r.round() as u64).unwrap_or(0),
+                address: String::new(),
+                confirmed: tx.confirmations > 0,
+                timestamp: tx.timestamp.unwrap_or(0),
```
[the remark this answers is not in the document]
coreyphillips: Good catch — confirmed the constraints (`onchain_activity.address` requires `length(address) > 0`, `activities.timestamp` requires `timestamp > 0`). Checked the apps: bitkit-android keeps these watch-only activities as in-memory display models (`StateFlow<List<Activity>>` rebuilt from watcher state) and never inserts them into the activity DB, which is why `address=""` is fine there. The original doc telling callers to `upsert_activity` was the bug. Reworded the doc to state these are display-only models not meant for DB insertion (242e151). Not fabricating a placeholder address/timestamp since that would diverge from how the apps render them.

src/modules/onchain/implementation.rs
```diff
@@ -1409,12 +1409,26 @@ pub(crate) fn map_bdk_tx_to_history(
         None => (None, None, 0),
     };
 
+    // Fee rate (sat/vB) requires both the fee and the raw transaction (for vsize).
+    let fee_rate = match (tx.fee, tx.transaction.as_ref()) {
+        (Some(fee), Some(raw_tx)) => {
+            let vsize = raw_tx.vsize();
+            if vsize > 0 {
+                Some(fee as f64 / vsize as f64)
```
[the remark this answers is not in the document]
coreyphillips: Correct, thank you — both `get_transaction_history` and the watcher's event builder called `list_transactions(false)`, so `tx.transaction` was `None` and `fee_rate` stayed `None` for every consumer. Switched both to `list_transactions(true)` (242e151) so vsize is available and the real fee rate is computed.

ovitrif:
Thanks, I hope the fact that we didn't share the full context why the 2 apps differ in implementation is not going to cause a lot of trouble to get this PR in its merge-able state.

The reason Android works more in-memory is mainly because it didn't have the later #103 PR at the time of the implementation.

The scope of that task actually surfaced as an idea during the Android implementation of Trezor subtask 1:
- Issue: https://github.com/synonymdev/bitkit-android/issues/1024
- PR: https://github.com/synonymdev/bitkit-android/pull/999

Since iOS had the primitives in place, Joao and myself synced to make sure iOS doesn't have to carry over this requirement to refactor later, which Android's first PR (ported by iOS' PR) embedded.

After #103 was merged, I updated the Android epic and its subtask (6) for optimisation with specs that came as a result of investigation how to levarage the walletId/wallet-scoped activities.

Links:
- Android Epic: https://github.com/synonymdev/bitkit-android/issues/998 (see info in issue description for subtask 6)
- Issue of Subtask 6: https://github.com/synonymdev/bitkit-android/issues/1029

ovitrif:
> The deterministic id = tx_id choice should be confirmed against how Bitkit already keys onchain activities; easy to switch to wallet_id:tx_id if cross-wallet collisions are a concern.

TxId might be problematic for cases where we send coins from Trezor to app's onchain wallet. My understanding is that activities for such actions will both have the same txId. This is one of the reasons we had to include more discriminators in the `walletId` PR #103

src/lib.rs
```diff
@@ -470,6 +470,120 @@ pub fn get_default_wallet_id() -> String {
...
+/// Android don't each hand-reconstruct activities from `HistoryTransaction` and drift.
+///
+/// A single hardware device often has several accounts/script-types, each watched
+/// separately, so the same txid can appear once per account (e.g. an internal
+/// transfer the device both sends and receives). Rows are therefore **merged by
+/// txid** — `received`/`sent` are summed and the tx is then classified as a whole —
+/// producing exactly one `Activity` per transaction. This mirrors the merge the
+/// mobile apps currently do by hand. Output preserves first-seen txid order.
+///
+/// This is a pure conversion — it does not touch the database. These are
```
ovitrif: I think this is solving the issue at the wrong boundary, app's are not really asking for the watcher to write into `ActivityDB` itself. The intended flow is that the watcher emits bitkit-core-compatible activity payloads, then Bitkit receives the watcher event and persists those through the existing Core activity APIs.

That means the emitted data still needs to be valid as normal onchain activity data. Hardware wallet transactions need to show in the same global activity list, sort together with other activities, be scoped by the hardware wallet `wallet_id`, support tags/contact/seen state, have transaction details, and survive backup/restore/rebuild flows. Returning display-only `Activity` values with `address = ""` and `timestamp = 0` makes that impossible because those rows cannot be inserted into the existing activity tables, effectively increasing the complexity for the app's integration instead of the desired outcome which is the opposite. App would have to carry these around, mix actual activities with a semi-baked Trezor-specific incompatible activity model and have to end up doing work that is meant to be shared across both platforms, since the logic for transforming these to activity DB-compatible structs will end up being duplicated across ios and android.

I think the watcher is the right place to build this payload because it still has the information the mapper has already lost. Before reducing everything to `HistoryTransaction`, the watcher has the BDK `Wallet`, raw transaction data from `list_transactions(true)`, confirmation time, fee/vsize, and ownership checks. `get_transaction_detail` already uses that kind of data to decode output addresses and mark wallet-owned outputs.

Could we change this so the watcher event/API emits persistence-ready activity data for the specific hardware wallet address type being watched? The implementation should populate the fields required by `activities` / `onchain_activity`, return matching `TransactionDetails`, use a positive timestamp for unconfirmed txs, and preserve user-owned fields like tags/contact/seen state on watcher refreshes.

src/lib.rs
```diff
@@ -470,6 +470,120 @@ pub fn get_default_wallet_id() -> String {
     DEFAULT_WALLET_ID.to_string()
 }
 
+/// Map a watch-only wallet's transaction history (as emitted by the xpub watcher
+/// in `WatcherEvent::TransactionsChanged`) into core `Activity` records, so iOS and
+/// Android don't each hand-reconstruct activities from `HistoryTransaction` and drift.
+///
+/// A single hardware device often has several accounts/script-types, each watched
```
ovitrif: Related to the watcher-emitted activity payload, I think the wallet boundary needs to stay at the specific hardware wallet address type being watched. If the same txid appears in two address types, those should be two wallet-scoped activities under different `wallet_id`s, not one merged device-level activity.

That lets Bitkit persist them through the normal activity APIs, show them together in the global activity list, and still scope them correctly when looking at one hardware wallet/address type.

Could we make the new watcher payload keep this boundary explicit instead of merging across address types by txid?

src/lib.rs
```diff
@@ -470,6 +470,120 @@ pub fn get_default_wallet_id() -> String {
...
+    }
+
+    order
+        .into_iter()
+        .map(|txid| {
+            let m = by_txid.remove(&txid).expect("txid present");
+            let (direction, amount, _net) = onchain::classify_tx(m.sent, m.received, m.fee);
+            let tx_type = match direction {
+                TxDirection::Received => PaymentType::Received,
+                TxDirection::Sent | TxDirection::SelfTransfer => PaymentType::Sent,
```
ovitrif: For whatever replaces this mapper, I think we need to preserve normal `OnchainActivity` value/fee semantics because Bitkit will render these through the same activity rows as every other onchain activity.

The current shape maps `SelfTransfer` to `Sent` with `value = fee` and `fee = fee`. Since Bitkit sent rows display `value + fee`, that would show double the actual cost if this shape becomes the persisted activity payload.

Could we make the watcher payload produce values that are compatible with the existing activity total calculation? For a pure self-transfer, I think that likely means `value = 0` and `fee = fee`, unless we choose another explicit Core representation.

src/lib.rs
```diff
@@ -470,6 +470,120 @@ pub fn get_default_wallet_id() -> String {
...
+                fee_rate: None,
+                timestamp: None,
+                confirmations: 0,
+            }
+        });
+
+        entry.received = entry.received.saturating_add(tx.received);
+        entry.sent = entry.sent.saturating_add(tx.sent);
+        // The same physical tx reports an identical fee/fee_rate from whichever
+        // account paid it; received-only rows report none. Keep the larger (real) one.
```
ovitrif: Same point for fees: the watcher payload should model fees from the watched wallet's perspective, not just copy whatever BDK attached to the history row. With raw tx enabled, BDK can provide `fee: Some(...)` for a receive-only wallet row, but that fee was paid by the sender, not this hardware wallet.

If Bitkit persists this as a normal received activity, storing/displaying that as the hardware wallet's fee would be wrong.

Could we derive `fee` / `fee_rate` based on whether the watched address-type wallet actually spent inputs, and add coverage for a receive-only row that still has `fee: Some(...)`?

bindings/ios/bitkitcore.swift
```diff
@@ -3694,6 +3698,9 @@ public struct HistoryTransaction {
         /**
          * Transaction fee in sats (None if not available, e.g. for received-only txs)
          */fee: UInt64?, 
+        /**
+         * Fee rate in sats per virtual byte (None if fee or tx size is unavailable).
+         */feeRate: Double?,
```
ovitrif, line 3703: I noticed they are not the same for all different binding targets. Let's make sure we always keep them in sync when we bump version and generate new bindings.

ovitrif (commented):
Leaving these together so the intended direction is clear. The watcher does not need to persist into the DB itself, but the watcher payload should be normal Core activity data that Bitkit can persist.

src/lib.rs
coreyphillips: Agreed, and done (540f894). Reworked it so the watcher builds persistence-ready Core activity data instead of a display-only model. `WatcherEvent::TransactionsChanged` now carries `activities: Vec<Activity>` + `transaction_details: Vec<TransactionDetails>`, built inside the watcher where the BDK `Wallet`, raw txs (`list_transactions(true)`), confirmation time, fee/vsize and `is_mine` are still available. Real decoded addresses, fees from the watched wallet's perspective, and a positive timestamp for unconfirmed txs mean the rows insert cleanly via the normal activity APIs. The standalone mapper is removed. I reused the decode path from `get_transaction_detail` (extracted into `map_bdk_tx_to_detail`) so it's shared.

src/lib.rs
coreyphillips: Done. Removed the cross-address-type merge. `WatcherParams` now takes a `wallet_id`, the watcher emits one activity per tx scoped to that wallet, and the same txid under two address types yields two wallet-scoped activities under different `wallet_id`s — the boundary stays at the watched address type.

src/lib.rs
coreyphillips: Fixed. Self-transfer is now `value = 0, fee = fee` so `value + fee` equals the true cost rather than double. Sent rows are `value = sent - received - fee, fee = fee`; received rows are `value = received, fee = 0`. Covered by tests `self_transfer_is_zero_value_plus_fee` and `sent_keeps_value_and_fee_separate`.

src/lib.rs
coreyphillips: Fixed. Fee/fee_rate are now attributed only when the watched wallet actually spent inputs (`tx.sent > 0`); a receive-only row that still carries BDK's `fee: Some(...)` (paid by the sender) maps to `fee = 0, fee_rate = 0`. Test `received_drops_sender_fee_and_uses_owned_address` exercises exactly that case.

bindings/ios/bitkitcore.swift
coreyphillips: Regenerated the Swift + Python bindings for this change. Note the per-target drift you spotted is from the repo convention where feature PRs only regen bitkitcore.swift + bitkitcore.py, and the iOS xcframework / Android Kotlin / committed dylib are rebuilt at the version bump via build.sh. Happy to also wire a CI check (or a build.sh --bindings-only step) that regenerates all targets together on version bump if you'd like to enforce sync.

coreyphillips:
@ovitrif thanks for the thorough review — this was exactly the right call. Pushed a redesign in 540f894 that moves activity-building into the watcher and emits persistence-ready `activities` + `transaction_details`, with per-address-type `wallet_id` scoping (no merge), correct value/fee semantics (self-transfer = value 0 + fee; receive-only drops the sender's fee), real decoded addresses, positive unconfirmed timestamps, and upsert hardening so refreshes preserve contact + first-seen timestamp. Replied inline on each point. Note this makes `WatcherEvent` / `WatcherParams` a breaking change vs the previous shape.

src/modules/activity/implementation.rs
```diff
@@ -1609,11 +1609,18 @@ impl ActivityDB {
...
+        //   refresh that doesn't know the contact can't wipe a user-set one.
+        // - `timestamp`: keep the existing value while the tx is still unconfirmed,
+        //   so a still-pending activity doesn't jump every refresh; snap to the
+        //   block time once confirmed.
         let activities_sql = "
             UPDATE activities SET
                 tx_type = ?1,
-                timestamp = ?2,
-                contact = ?3
+                timestamp = CASE WHEN ?6 THEN ?2 ELSE timestamp END,
```
ben-kaufman: I think the batch upsert path still misses this preservation. `upsert_activities` sends watcher `Vec<Activity>` values through `upsert_onchain_activities`, and that conflict clause still replaces `timestamp` and `contact`, so a refresh can clear a saved contact and move a pending tx's first-seen time. Could we apply the same unconfirmed timestamp/contact preservation there too?

ben-kaufman: This also changes explicit `update_activity` calls, not only watcher refreshes. For an unconfirmed onchain activity, callers can no longer update `timestamp`, and passing `contact = None` can no longer clear an existing contact. Could we keep `update_activity` as a real replacement and move the watcher preservation into the upsert path?

ben-kaufman (commented):
Review comments from local pass.

src/modules/onchain/implementation.rs
```diff
@@ -1433,6 +1450,214 @@ pub(crate) fn sort_history_transactions(history: &mut [HistoryTransaction]) {
...
+        contact: None,
+        created_at: None,
+        updated_at: None,
+        seen_at: None,
+    });
+
+    let details = TransactionDetails {
+        wallet_id: wallet_id.to_string(),
+        tx_id: detail.txid.clone(),
+        amount_sats: detail.net,
```
ben-kaufman: `TransactionDetails.amount_sats` is documented as excluding fees, but this stores `detail.net`, which is `received - sent`. For sent and self-transfer rows, BDK's `sent` includes the fee, so the persisted detail amount becomes fee-inclusive while the activity value excludes the fee. Could we store the fee-excluded amount here, or update the details contract if fee-inclusive is intended?

src/modules/activity/implementation.rs
coreyphillips: Good catch, fixed in 9ec312d. The preservation now lives in the batch upsert: `upsert_onchain_activities` ON CONFLICT keeps the existing `contact` when the incoming one is NULL (`COALESCE`) and keeps the existing `timestamp` while the tx is unconfirmed (`CASE WHEN ?confirmed THEN excluded.timestamp ELSE timestamp END`), snapping to the block time on confirmation. Test: `test_batch_upsert_preserves_contact_and_unconfirmed_timestamp`.

coreyphillips: Agreed, reverted in 9ec312d. `update_onchain_activity_by_id` is a literal replacement again, so `update_activity` can clear a contact (`None`) and move an unconfirmed tx's timestamp. The preservation moved entirely into the upsert path, and singular `upsert_activity` now routes through the batch `upsert_onchain_activities` so both upsert entry points behave identically. Test: `test_update_activity_is_pure_replacement`.

src/modules/onchain/implementation.rs
coreyphillips: Fixed in 9ec312d. `amount_sats` is now fee-excluded to match its doc contract: since BDK's `sent` includes the fee, I add the wallet's own fee back (`net + fee`). Receive-only rows are unaffected because their perspective fee is 0. e.g. sent 60000 (incl 500 fee), received 0 gives amount_sats = -60000 + 500 = -59500, matching the 59500 that actually left. Asserted in the builder tests.

coreyphillips:
@ovitrif on the `tx_id` point (a Trezor send to the app's own onchain wallet producing the same txid in both): the redesign keys activities on the composite primary key `(wallet_id, id)` with `id = tx_id`, and each watcher is scoped to a specific address-type `wallet_id`. So the Trezor-side activity (`wallet_id = trezor:...`) and the app-wallet-side activity (its own `wallet_id`) are distinct rows even with the same txid, no collision. That matches the wallet-scoping direction from #103. Thanks also for the context on why Android went in-memory (pre-#103) and the epic/subtask links, that lines up with this approach.

ben-kaufman (approved):
Full review looks good. I rechecked the watcher/activity flow, generated surfaces, and the resolved threads.

ovitrif:
LGTM, reviewed after merge. Thanks everyone for the help 🙏🏻

PR #111, opened 2026-06-25
feat: core-owned supported hardware-wallet catalog (#107 subtask 4)

### LINKED ISSUE #107: Updates to facilitate Trezor integration in mobile

coreyphillips:
Closes #107

## Summary

Subtask 4 of #107. Core now owns the catalog of supported hardware wallets, so iOS and Android render the supported-device list generically instead of each hardcoding Trezor metadata. Follow-up from the review note on synonymdev/bitkit-android#1039 ("this could come from Bitkit-core in the future, because it is where all supported HW info is actually located").

## What changed

New UniFFI surface in the trezor module (re-exported from `lib.rs`):

```rust
pub enum HardwareWalletVendor { Trezor }   // extensible to other vendors

pub struct SupportedHardwareWallet {
    vendor: HardwareWalletVendor,
    vendor_name: String,        // "Trezor"
    model: String,              // "Safe 7"  (stable key for image mapping)
    display_name: String,       // "Trezor Safe 7"
    transports: Vec<TrezorTransportType>,
}

pub fn get_supported_hardware_wallets() -> Vec<SupportedHardwareWallet>;
```

Catalog (Trezor's full lineup):

| display_name | transports |
|---|---|
| Trezor Model One | Usb |
| Trezor Model T | Usb |
| Trezor Safe 3 | Usb |
| Trezor Safe 5 | Usb |
| Trezor Safe 7 | Usb, Bluetooth |

## Platform handling

Core stays platform-agnostic. Apps filter on `transports`:
- **Android** supports all models, renders the whole catalog.
- **iOS** is Bluetooth-only, so it shows only entries whose `transports` include `Bluetooth` (currently just the Safe 7).

## Notes

- Fields are intentionally minimal. Runtime capabilities (passphrase entry, etc.) already come from `TrezorFeatures` on a connected device, so the catalog only needs identity + transports. Apps map `model` to their own bundled images.
- Version bumped to **0.3.5**; all platform bindings + native artifacts regenerated via `build.sh all`, and the `Package.swift` checksum matches the rebuilt `BitkitCore.xcframework.zip`.

## Tests

`cargo test --lib supported_hardware_wallets` — asserts the lineup, well-formed entries (USB on every model, `display_name` = `"Trezor {model}"`), and that the Bluetooth-capable filter yields exactly the Safe 7.


ovitrif (approved):
LGTM, thanks 🙏🏻

PR #115, opened 2026-06-25
fix: validate LNURL-pay amount

ovitrif:
### Description

This PR:
1. Adds an LNURL-pay callback invoice helper for scanned LNURL-pay data.
2. Validates that returned LNURL-pay invoices include an amount.
3. Rejects returned LNURL-pay invoices when the invoice amount differs from the requested amount.

### Preview

N/A

#### Manual Tests

N/A

#### Automated Checks

- `cargo fmt --check`
- `cargo test lnurl`
- `git diff --check`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -63,7 +63,7 @@ use crate::onchain::{
 pub use modules::activity;
 pub use modules::lnurl;
 pub use modules::onchain;
-pub use modules::scanner::{DecodingError, Scanner};
+pub use modules::scanner::{DecodingError, LnurlPayData, Scanner};
 
 use bip39::Mnemonic;
 use bitcoin::bip32::Xpriv;
@@ -115,6 +115,20 @@ pub async fn get_lnurl_invoice(
     rt.spawn(async move { lnurl::get_lnurl_invoice(&address, amount_satoshis).await })
         .await
         .unwrap()
+}
+
+#[uniffi::export]
+pub async fn get_lnurl_invoice_for_pay_data(
+    data: LnurlPayData,
+    amount_msats: u64,
+    comment: Option<String>,
+) -> Result<String, lnurl::LnurlError> {
+    let rt = ensure_runtime();
+    rt.spawn(
+        async move { lnurl::get_lnurl_invoice_for_pay_data(data, amount_msats, comment).await },
+    )
+    .await
+    .unwrap()
 }
 
 #[uniffi::export]
--- a/src/modules/lnurl/implementation.rs
+++ b/src/modules/lnurl/implementation.rs
@@ -1,11 +1,19 @@
 use crate::lnurl::{ChannelRequestParams, LnurlAuthParams, LnurlError, WithdrawCallbackParams};
+use crate::modules::scanner::LnurlPayData;
 use bitcoin::bip32::Xpriv;
 use bitcoin::secp256k1::{Message, PublicKey, Secp256k1};
+use lightning_invoice::Bolt11Invoice;
 use lnurl::lightning_address::LightningAddress;
 use lnurl::lnurl::LnUrl;
 use lnurl::{get_derivation_path, AsyncClient, Builder, LnUrlResponse, Response};
+use serde::Deserialize;
 use std::str::FromStr;
 use url::Url;
+
+#[derive(Deserialize)]
+struct LnurlPayCallbackResponse {
+    pr: Option<String>,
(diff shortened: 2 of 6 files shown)


src/lib.rs
```diff
@@ -115,6 +115,20 @@ pub async fn get_lnurl_invoice(
         .unwrap()
 }
 
+#[uniffi::export]
+pub async fn get_lnurl_invoice_for_pay_data(
+    data: LnurlPayData,
+    amount_msats: u64,
+    comment: Option<String>,
+) -> Result<String, lnurl::LnurlError> {
```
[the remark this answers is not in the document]
ovitrif: We don't care about Python bindings; they'll be removed soon.

coreyphillips (approved):

src/modules/lnurl/implementation.rs
```diff
@@ -23,6 +32,33 @@ pub async fn get_lnurl_invoice(address: &str, amount_satoshis: u64) -> Result<St
...
 
+pub async fn get_lnurl_invoice_for_pay_data(
+    data: LnurlPayData,
+    amount_msats: u64,
+    comment: Option<String>,
+) -> Result<String, LnurlError> {
+    validate_amount_msats(amount_msats, data.min_sendable, data.max_sendable)?;
+
+    let callback_url =
+        build_lnurl_pay_callback_url(&data.callback, amount_msats, comment.as_deref())?;
```
ben-kaufman, line 42: Could we validate `comment` against `data.comment_allowed` before building the callback URL and add a regression test for an over-limit comment? `LnurlPayData` already carries the server's max length, but this path currently forwards any non-empty comment to the callback.

ovitrif: Will do as part of post-release merge hygiene, which will also require merging locally with master, changing PR branch target base, and new version on top of merged changes 🙏🏻 

thanks 🫡

ovitrif: Agreed this is worth tracking after the release branch work. For this patch, I’m keeping the change limited to returned invoice amount validation and leaving comment length handling for follow-up release hygiene.

Resolved in 496c084

coreyphillips:
LGTM!

Only small items that popped up were the following:

### 1. Dead code from a checked-in reversal

Commit 1 added a LUD-06 description-hash / metadata check; commit 2 removed it. The result is:

* `validate_lnurl_pay_invoice`'s `_metadata` parameter is now unused, and `data.metadata_str` is passed through for no reason.
* The public `LnurlError::MetadataMismatch` variant is now dead API surface in the generated Swift, Kotlin, and Python bindings.
* The renamed test `..._with_different_description` now documents that metadata is intentionally not checked.

### 2. No HTTP timeout on the callback fetch

The callback uses `reqwest::get`, which creates a client with no request timeout and follows redirects. A malicious or slow callback server can therefore keep the spawned task hanging indefinitely (e.g. via a Slowloris-style response).

coreyphillips (approved):

ovitrif:
Thanks @coreyphillips, reason to keep API surface changes smallest was for temporary reasons to not invalidate too much of previous session's testing; more cleanup will follow on this PR before merging it back to master 🙏🏻 

Then we can change apps too so they don't need handling of "dead code" like `MetadataMismatch` error types, and those can go through normal release procedure which allows time to test.

ovitrif:
@ben-kaufman Context on why the validation was narrowed down so much:
Noticed a blocking regression to the patch release during pre-release testing: WoS payments started failing because metadata was returned in unexpected format which didn't adhere to the protocol expectations, concluded that it's better not to validate anything besides amount to reduce impact surface of potential regressions, after all, only amount was critical validation for security => ended up not validating anything else, which is why your suggested comments validation was also descoped.

Thanks 🙏🏻

ovitrif:
Current status after the rework:

- This PR is merged into `master` as `f830b59`.
- The merged diff is source-only; generated bindings, binary artifacts, and version-release noise were dropped.
- `v0.3.7` was published before this merge, so it does not include this fix.
- The `v0.3.7` Android publish workflow failed, so the Android package/native-symbol upload is incomplete anyway.
- `synonymdev/bitkit-core#118` is still open and only contains iOS binding artifacts from before this PR merged.

Follow-up: cut a fresh release that includes both `synonymdev/bitkit-core#117` and this PR, and verify the Android publish workflow succeeds.

PR #116, opened 2026-06-26
feat(boltz): Boltz submarine & reverse swaps

coreyphillips:
## Summary

Adds a `boltz` module integrating [Boltz](https://boltz.exchange) **submarine** (onchain → Lightning) and **reverse** (Lightning → onchain) swaps behind the UniFFI surface, for iOS/Android/Python.

The dangerous cryptography (MuSig2 Taproot cooperative signing, swap scripts, claim/refund tx construction) is delegated to the `boltz-client` crate. This module adds deterministic key management, SQLite persistence, lifecycle tracking, automatic claiming, and the FFI surface.

## Key design decision: deterministic keys, no stored secrets

Swap keys and reverse-swap preimages are **derived from the wallet seed** via Boltz's BIP85 scheme (`SwapMasterKey`/`derive_swapkey`, `Preimage::from_swap_key`) — never random, never persisted. `boltz.db` stores only a **monotonic per-swap derivation index**.

Consequences:
- A leaked database cannot move funds (it holds no key material).
- Swaps are recoverable two ways: same-device (index + in-memory seed), or **seed-only** via Boltz's rescue API if `boltz.db` is lost.
- The wallet `mnemonic` (+ optional BIP39 passphrase) now flows through the create/claim/refund/start-updates FFI calls. The background updates stream holds the mnemonic in memory only for its lifetime (dropped on stop) to auto-claim. **The passphrase must match the wallet's**, or derived keys won't control the funds.

> Reviewers: please confirm threading the seed through these entry points fits Bitkit's key-handling conventions on the Swift/Kotlin side.

## What's included

- Submarine create/refund and reverse create/claim, with cooperative key-path → script-path fallback.
- Managed WebSocket updates stream; auto-claims reverse swaps on `transaction.confirmed` (not mempool, to avoid revealing the preimage against an unconfirmed lockup).
- Atomic, collision-free swap-index reservation; `PRAGMA user_version` migration anchor; input validation on create.
- Idempotent claim/refund — returns the recorded txid without re-broadcasting.
- Typed lifecycle status with forward-compatible `Unknown { raw }`; recovery/listing APIs.
- Only one updates stream (one network) runs at a time — documented.

## Testing

`cargo build`, all 9 boltz unit tests, clippy, and fmt are clean. Tests cover status mapping, DB round-trip/recovery, monotonic index reservation, and deterministic derivation. An ignored live E2E test creates a real reverse swap and cryptographically validates the locally-derived redeem script + invoice against Boltz's response (no broadcast).

**Known follow-up:** the claim/refund *broadcast* paths are not yet covered by an automated test — they need a regtest Boltz + Electrum stack. Recommended as a follow-up.


src/modules/boltz/api.rs
```diff
@@ -0,0 +1,224 @@
...
+        let request = CreateSubmarineRequest {
+            from: "BTC".to_string(),
+            to: "BTC".to_string(),
+            invoice: invoice.clone(),
+            refund_public_key,
+            pair_hash: None,
+            referral_id: None,
+            webhook: None,
+        };
+        let response = client
```
ovitrif, line 94: Could we validate the Boltz create response before persisting or returning it? `boltz-client` exposes `response.validate(...)` for both submarine and reverse swaps, and that check proves the returned script/address matches our derived key and invoice/preimage before the app sends funds or pays the invoice.

src/lib.rs
```diff
@@ -2698,3 +2711,263 @@ pub fn onchain_stop_watcher(watcher_id: String) -> Result<(), AccountInfoError>
...
+
+fn boltz_runtime_err(e: tokio::task::JoinError) -> BoltzError {
+    BoltzError::ConnectionError {
+        error_details: format!("Runtime error: {}", e),
+    }
+}
+
+/// Fetch fees and limits for submarine swaps (onchain -> Lightning).
+#[uniffi::export]
+pub async fn boltz_get_submarine_limits(
```
ovitrif, line 2754: Since this adds new UniFFI exports and public types, could we bump the package version and regenerate the mobile/Python bindings in this PR? Right now Rust builds, but downstream iOS/Android consumers would not get the new `boltz*` APIs from the checked-in artifacts.

src/lib.rs
```diff
@@ -2698,3 +2711,263 @@ pub fn onchain_stop_watcher(watcher_id: String) -> Result<(), AccountInfoError>
...
+) -> Result<String, BoltzError> {
+    let rt = ensure_runtime();
+    rt.spawn(async move {
+        let db = get_boltz_db().await?;
+        let record = db.get_swap(&swap_id).await?.ok_or(BoltzError::NotFound {
+            error_details: format!("Swap {} not found", swap_id),
+        })?;
+        // Idempotent: don't re-broadcast a swap that already has a claim tx
+        // (e.g. an auto-claim that ran first).
+        if let Some(existing) = record.claim_tx_id.clone() {
```
ovitrif: Could we serialize claim/refund attempts per swap, or mark a claim/refund as in progress before broadcasting? As written, auto-claim and a manual recovery call can both see no recorded txid and try to broadcast before either path persists the result.

ovitrif (commented):
reviewed the current draft, added a few considerations in scoped review comments

src/modules/boltz/api.rs
coreyphillips: Good catch, done. Both create paths now validate before the response can do any damage.

`create_submarine_swap` calls `response.validate(&invoice, &refund_public_key, network.as_chain())` and `create_reverse_swap` calls `response.validate(&preimage, &claim_public_key, network.as_chain())`. Both run immediately after the API call and **before** `insert_swap` and before the response is handed back, so a response that fails validation never reaches the database and never reaches the app. That ordering is the point: for submarine swaps the caller is about to fund the lockup address, and for reverse swaps it is about to pay the invoice, so validating after persisting would still let the app act on a bad response.

Failures surface as `BoltzError::SwapError` with the underlying reason.

One limitation worth stating plainly: I could not exercise this against a live Boltz response. The `live_reverse_swap` test skips right now because the Boltz testnet API is returning 502 Bad Gateway, so the validation is covered by the type checker and the existing offline tests, not by an observed round trip against a real server. Once testnet is healthy, `cargo test modules::boltz -- --ignored` will cover it.

src/lib.rs
coreyphillips: You are right, and the idempotency check I had was not sufficient. Reading `claim_tx_id`, broadcasting, then recording the txid spans several `.await` points (Boltz plus Electrum round trips), so the auto-claim and a manual recovery call could both observe `None` and both broadcast. Checking harder at the top could not fix that; the check and the write had to become atomic.

Fixed by serializing per swap. New `src/modules/boltz/guard.rs` keeps a per-swap-id async lock, and claiming and refunding now go through single guarded entry points, `claim_reverse_swap_guarded` and `refund_submarine_swap_guarded`. Each one takes the swap's lock, **re-reads the record under the lock** (the caller's copy may be stale by then), returns the recorded txid if one appeared while it waited, and otherwise broadcasts and persists while still holding the lock.

Both racing paths now funnel through that: `boltz_claim_reverse_swap` / `boltz_refund_submarine_swap` in `lib.rs`, and `auto_claim` in the listener. Whichever arrives second waits, sees the first one's txid, and broadcasts nothing. The claim path returns a `ClaimOutcome` so the listener can tell the two cases apart: it only emits `BoltzSwapEvent::Claimed` when it actually broadcast, rather than double-reporting a claim the manual caller already received.

Tests added:
- a claim of an already-claimed swap returns the recorded txid without broadcasting (the record points at an unreachable Electrum server, so reaching the broadcast path at all would fail the test rather than pass it)
- the same for refunds
- a missing swap reports `NotFound`
- four concurrent claims of one swap never overlap in the critical section, while locks on different swaps stay independent

Scope: this serializes within the process, which is where the race you identified lives, since both racers are in this library. It is not a cross-process lock.

src/lib.rs
coreyphillips: Done. Bumped to **0.5.0** and regenerated the iOS, Android and Python bindings, so the `boltz*` APIs are now in the checked-in artifacts.

Minor rather than patch: this PR adds new public API, and `build.sh`'s default patch bump would have collided with the 0.4.1 already sitting on `fix/broadcast-return-known-txid`.

Version is in sync across `Cargo.toml`, `Package.swift`, `bindings/android/gradle.properties` and `bindings/python/setup.py`. The generated surfaces now carry the boltz exports, where before they had none: 231 references in `bitkitcore.swift`, 285 in `bitkitcore.android.kt`, 429 in the Python module, plus rebuilt native libraries for all four Android ABIs and both iOS slices.

**One thing that needs a second opinion, since it changes what the repo tracks.** Linking `boltz-client` grew the xcframework's static libraries from ~98 MB to ~113 MB, which is over GitHub's hard 100 MB per-file limit, so the push was rejected outright. They were already within 2 MB of the limit on master, so this was going to happen on the next dependency regardless. Stripping only got them to 103 MB, still over.

I stopped tracking `BitkitCore.xcframework/**/libbitkitcore.a` and gitignored it. Those files only duplicate the compressed copies inside `BitkitCore.xcframework.zip` (76 MB, still committed), and that zip is what `Package.swift` actually downloads from the GitHub release via `binaryTarget(url:checksum:)`, so no consumer resolves iOS through the expanded directory. Nothing about the SPM consumption path changes.

Worth deciding as a team whether that is the fix you want long term, or whether these binaries should move to Git LFS or out of the repo entirely. The zip is at 76 MB and climbing, so it will hit the same wall eventually.

Separately: `build.sh -r --<bump> all` fails at the Android step if `ANDROID_NDK_HOME` points at an NDK that is not installed, and on the way out it *deletes* the checked-in `jniLibs` and Kotlin bindings instead of leaving them alone. It also strips `[[bin]] example` from `Cargo.toml` and removes `example/main.rs`, and skips its own cleanup if interrupted. I restored all of that by hand here, but the script could use some hardening.

coreyphillips:
Latest release for testing can be found as v0.5.1 [here](https://github.com/synonymdev/bitkit-core/releases/tag/v0.5.1).

src/modules/boltz/api.rs
```diff
@@ -0,0 +1,246 @@
...
+            .post_swap_req(&request)
+            .await
+            .map_err(map_api_err("create submarine swap"))?;
+
+        // Prove Boltz's response before persisting it or handing the caller a
+        // lockup address to fund: this rebuilds the swap script from our own
+        // refund key and the invoice's preimage hash and checks it hashes to the
+        // address Boltz returned. Without it, a malicious or buggy server could
+        // hand back an address we cannot refund from.
+        response
```
ovitrif: I think the current dependency validation leaves several Bitcoin response terms unbound: submarine validation does not compare the script hashlock with the invoice, while reverse validation does not compare the script hashlock with the preimage or the invoice amount with `amount_sat`. Could we add those checks before returning payment details, with focused mismatched-response tests?

src/modules/boltz/db.rs
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    /// List swaps that have not reached a terminal state, for recovery and for
+    /// resubscribing to status updates after a restart.
+    pub async fn list_pending_swaps(&self) -> Result<Vec<SwapRecord>, BoltzError> {
+        Ok(self
+            .list_swaps()
+            .await?
+            .into_iter()
+            .filter(|r| !BoltzSwapStatus::from_raw(&r.status).is_terminal())
```
ovitrif: I think recovery needs to depend on local completion state here. A cooperative reverse claim can disclose the preimage before Electrum broadcast; if broadcast then fails, Boltz may report `invoice.settled` and this filter removes the swap even though `claim_tx_id` is absent. Delayed server updates can also overwrite a locally recorded completion status. Could we keep the swap recoverable while its local txid is absent and preserve local completion once a txid exists?

src/modules/boltz/listener.rs
```diff
@@ -0,0 +1,380 @@
...
+                    Ok(status) => process_status(&db, listener.as_ref(), &config, status).await,
+                    Err(RecvError::Lagged(_)) => continue,
+                    Err(RecvError::Closed) => break,
+                }
+            }
+        })
+    };
+
+    let mut guard = updates_cell().lock().await;
+    *guard = Some(UpdatesHandle {
```
ovitrif, line 217: I think we need to serialize the updates-stream replacement. Two concurrent starts can both pass `stop_swap_updates`, spawn their tasks, and then replace each other here. Dropping the first handle does not abort its detached tasks, so an obsolete stream can remain active. Could we serialize the start/stop transition and ensure tasks spawned during failed or superseded initialization are aborted, with a concurrent-start test?

src/modules/boltz/api.rs
```diff
@@ -0,0 +1,246 @@
...
+        claim_address: String,
+        mnemonic: String,
+        bip39_passphrase: Option<String>,
+    ) -> Result<ReverseSwapResponse, BoltzError> {
+        if amount_sat == 0 {
+            return Err(BoltzError::InvalidInput {
+                error_details: "amount_sat must be greater than 0".to_string(),
+            });
+        }
+        if claim_address.trim().is_empty() {
```
ovitrif: Could we validate `claim_address` against the selected network before sending the create request? Reverse creation currently only checks that it is nonblank, while parsing and network validation happen later, after the caller may have paid the invoice, and the stored destination cannot be replaced.

src/modules/boltz/claim.rs
```diff
@@ -0,0 +1,126 @@
...
+        .clone()
+        .ok_or_else(|| BoltzError::InvalidInput {
+            error_details: "Reverse swap is missing a claim address".to_string(),
+        })?;
+
+    let chain = record.network.as_chain();
+    let swap_script = SwapScript::reverse_from_swap_resp(chain, &reverse_resp, our_pubkey)?;
+    let chain_client = build_chain_client(record.network, &record.electrum_url)?;
+    let boltz_client = build_boltz_client(record.network);
+    let fee = Fee::Relative(fee_rate_sat_per_vb.unwrap_or(DEFAULT_FEERATE_SAT_PER_VB));
```
ovitrif, line 93: Could we require a finite, positive fee rate before starting updates or constructing a claim/refund? Arbitrary `f64` values reach `Fee::Relative`; negative values and `NaN` become zero, while infinity saturates, so invalid configuration can fail only after fund-moving work begins.

src/modules/boltz/README.md
```diff
@@ -0,0 +1,308 @@
...
+boltzGetSubmarineLimits(network)                         -> BoltzPairInfo
+boltzGetReverseLimits(network)                           -> BoltzPairInfo
+boltzCreateSubmarineSwap(network, electrumUrl, invoice, mnemonic, bip39Passphrase?)         -> SubmarineSwapResponse
+boltzCreateReverseSwap(network, electrumUrl, amountSat, claimAddress, mnemonic, bip39Passphrase?) -> ReverseSwapResponse
+boltzListSwaps()                                         -> [BoltzSwap]
+boltzListPendingSwaps()                                  -> [BoltzSwap]
+boltzGetSwap(swapId)                                     -> BoltzSwap?
+boltzClaimReverseSwap(swapId, mnemonic, bip39Passphrase?, feeRateSatPerVb?)          -> String (txid)
+boltzRefundSubmarineSwap(swapId, refundAddress, mnemonic, bip39Passphrase?, feeRateSatPerVb?) -> String (txid)
+boltzStartSwapUpdates(network, listener, mnemonic, bip39Passphrase?, feeRateSatPerVb?) // managed WebSocket
```
ovitrif: Could we add the explicit `acceptZeroConf` argument to the API table and examples, using `false` in the safety-first examples? The generated Swift, Kotlin, and Python APIs require it, while the documentation still uses the five-argument signature.

ovitrif (commented):
I focused this review on the Boltz feature behavior and its fund-recovery paths. I left six scoped comments covering response validation, recovery state, updates lifecycle, input validation, and the documented FFI signature.

src/modules/boltz/api.rs
coreyphillips: Confirmed, and worth spelling out how much validate() leaves open on the Bitcoin branch: CreateSubmarineResponse::validate parses the invoice but never compares the script hashlock to it (only the Liquid arm does), and CreateReverseResponse::validate checks the invoice's payment hash and the lockup address but neither the script hashlock nor the invoice amount. A server could return an internally consistent script/address pair whose hashlock it controls, and for submarine swaps that is a theft vector: it could claim the lockup with its own preimage without paying the invoice. For reverse swaps a wrong hashlock silently breaks our script-path claim, and an unchecked amount lets the invoice ask for more than requested.

Fixed in 498be8f with a new validation module. validate_submarine_response binds the script hashlock to the invoice's payment hash160, and validate_reverse_response binds the script hashlock to our preimage's hash160 and the invoice amount to the requested amount_sat. Both wrap the existing response.validate call and run before anything is persisted or returned.

The mismatched-response tests build fixtures whose script, tree and taproot lockup address all agree with each other, so boltz-client's own address validation passes and only the term under test is wrong, which is exactly the malicious-server shape. Covered: reverse hashlock mismatch, reverse invoice amount mismatch, submarine hashlock mismatch, plus consistent baselines for both directions that must validate.

src/modules/boltz/db.rs
coreyphillips: You are right on both halves, and the scenario is real: the cooperative claim posts the preimage to Boltz before the Electrum broadcast, so Boltz can settle the invoice while our claim transaction never reached the chain. With the old filter that swap silently left the pending set with funds still sitting in the lockup.

Fixed in 498be8f. Recovery now depends on local completion state via SwapRecord::is_locally_complete: a terminal server status only completes a reverse swap if a claim txid is recorded locally, so invoice.settled with no local txid stays in the pending set. And update_status now refuses to touch a row once claim_tx_id or refund_tx_id is set, so a delayed or re-ordered server update (the WebSocket and the reconcile loop both feed it) cannot overwrite a locally recorded completion.

One step further so the swap actually completes rather than just staying visible: should_auto_claim now also fires on invoice.settled when no local claim txid exists. The periodic reconcile therefore retries the broadcast every cycle until it lands. The preimage is already disclosed at that point, so retrying carries no added risk.

Tests added: a settled reverse swap without a local claim txid stays recoverable and drops out once set_claim_tx records one; late server updates cannot regress a locally recorded claimed/refunded status; and the new invoice.settled arm of should_auto_claim, including that a recorded txid still suppresses it.

src/modules/boltz/listener.rs
coreyphillips: Confirmed. Dropping a JoinHandle detaches rather than aborts, so two concurrent starts could both pass the stop, and whichever handle was overwritten left its three tasks (WebSocket loop, processor, reconciler) running forever. The error path had the same leak in miniature: a failing list_pending_swaps returned after the WebSocket task was already spawned.

Fixed in 498be8f by serializing the whole transition. start_swap_updates now holds the stream slot's mutex across teardown, setup and publication: it takes the old handle and calls its new shutdown() (which aborts all three tasks and drops the WebSocket), runs the only fallible step before anything is spawned so a failed start cannot leak a task, then spawns and stores the new handle while still holding the lock. stop_swap_updates goes through the same shutdown path, and the old internal stop call inside start is gone (it would now deadlock, which is a nice guarantee that every transition goes through the one lock).

For the test, aborts had to be observable, so each spawned task future carries an RAII marker counted in a static gauge; the count drops when the future is dropped, whether it completed or was aborted. The new test runs four concurrent starts against a shared slot and asserts exactly one stream (three live tasks) survives, then that stop_swap_updates takes it to zero.

src/modules/boltz/api.rs
coreyphillips: Done in 498be8f. create_reverse_swap now runs the claim address through validate_onchain_address before the create request: it must parse as a Bitcoin address and belong to the selected network, and the canonical (trimmed) form is what gets persisted as the claim destination. A malformed or wrong-network address now fails at creation, before any invoice exists to pay, instead of at claim time when the destination can no longer be replaced.

The refund path gets the same treatment for consistency: refund_submarine_swap_guarded validates the refund address against the swap's network before any transaction construction. That check sits after the idempotent early return, since a swap with a recorded refund txid returns it without using the address at all.

Tests cover a mainnet address rejected on testnet and regtest (and the reverse), garbage and empty input, and that surrounding whitespace is tolerated and stripped. The live E2E test now also picks its sample claim address per network so it keeps working under BOLTZ_LIVE_NETWORK overrides.

src/modules/boltz/claim.rs
coreyphillips: Done in 498be8f. validate_fee_rate requires a provided rate to be finite and strictly positive (None still falls back to the built-in default), and it runs at every entry point that leads to Fee::Relative: boltz_start_swap_updates rejects a bad auto-claim rate before any stream state exists, and claim_reverse_swap_guarded and refund_submarine_swap_guarded reject one at the top, before the swap lock is taken or any Boltz/Electrum work begins. Invalid configuration now surfaces as InvalidInput immediately instead of after fund-moving work has started.

Tests cover NaN, negative, zero and both infinities being rejected, and that valid rates and None pass.

src/modules/boltz/README.md
coreyphillips: Done in 498be8f. The API table now shows the six-argument boltzStartSwapUpdates signature including acceptZeroConf, and every example that starts the stream (Swift, Kotlin, Python, and the recovery-after-restart snippet) passes it explicitly as false with a note that false claims only after the lockup confirms. The lifecycle section now describes acceptZeroConf as the opt-in for mempool-time claiming instead of the outdated suggestion to call boltzClaimReverseSwap manually for the 0-conf case, and the recovery section was updated for the new local-completion semantics of boltzListPendingSwaps from the other thread.

src/modules/boltz/refund.rs
```diff
@@ -0,0 +1,108 @@
...
+
+    let txid = refund_submarine_swap(
+        &record,
+        refund_address,
+        mnemonic,
+        bip39_passphrase,
+        fee_rate_sat_per_vb,
+    )
+    .await?;
+    db.set_refund_tx(swap_id, &txid).await?;
```
ovitrif: nit: I think the validated `refund_address` should be persisted here. After a successful refund, `boltzGetSwap` and `boltzListSwaps` still return a null `onchainAddress`, even though that field represents the submarine refund destination. Could we store the canonical address with the completion update and cover the round-trip?

src/modules/boltz/listener.rs
```diff
@@ -0,0 +1,526 @@
...
+                    false,
+                ))
+            })
+            .collect();
+        for start in starts {
+            start.await.unwrap().unwrap();
+        }
+
+        // All but the last-written stream must have been fully aborted.
+        wait_for_live_tasks(3);
```
ovitrif: nit: I think both `wait_for_live_tasks(...)` calls need `.await`. They currently create and immediately drop futures, so the test would still pass if superseded or stopped tasks leaked. Could we await both checks so the test exercises the shutdown guarantee?

ovitrif (approved):
Looks good overall. I left two non-blocking nits around refund metadata persistence and the stream-lifecycle test.

src/modules/boltz/refund.rs
coreyphillips: Good catch, addressed in 413aedb. `set_refund_tx` now takes the validated refund address and persists it as the swap's `onchain_address` together with the txid and terminal status, so `boltzGetSwap` and `boltzListSwaps` return the refund destination after completion. The db test now covers the round-trip.

src/modules/boltz/listener.rs
coreyphillips: Fixed in 413aedb. Both `wait_for_live_tasks` calls are now awaited, so the checks actually run and the test fails if superseded or stopped stream tasks leak.

coreyphillips:
New build and version bump incoming...

coreyphillips:
Updated build/bindings [here](https://github.com/synonymdev/bitkit-core/releases/tag/v0.5.4).

Issue #113, opened 2026-06-25
fix: add wallet-scoped backup migrations


ovitrif:
## Problem

Android and iOS backup restore currently decode app-owned backup JSON directly into generated Core model types before calling Core.

That is brittle for wallet-scoped activity changes. If a backup was written before `walletId` existed, the app can fail while decoding `Activity`, `ActivityTags`, or `PreActivityMetadata`, before Core gets a chance to apply defaults or migrate the data.

Current app restore paths:

- Android `ActivityBackupV1` contains `activities: List<Activity>`, `activityTags: List<ActivityTags>`, and `closedChannels: List<ClosedChannelDetails>`. `BackupRepo` decodes that payload, then `ActivityRepo.restoreFromBackup()` calls Core `upsertActivities`, `upsertTags`, and `upsertClosedChannels`.
- Android `MetadataBackupV1` contains `tagMetadata: List<PreActivityMetadata>`. `BackupRepo` decodes that payload, then upserts the metadata through Core.
- iOS has the same backup shapes in `ActivityBackupV1` and `MetadataBackupV1`. `BackupService` decodes them with `JSONDecoder`, then calls Core `upsertList`, `upsertTags`, `upsertClosedChannelList`, and `upsertPreActivityMetadata`.
- Android's normal wallet id is `bitkit` via `ActivityWalletType.BITKIT.id`; Trezor wallet ids are `trezor:{sha256(sorted xpubs joined by \n)}`.
- iOS uses `WalletScope.default = getDefaultWalletId()` for the normal wallet; Trezor wallet ids are derived in `HwWalletId` with the same `trezor:{sha256(sorted xpubs joined by \n)}` shape when xpubs are available.
- Both apps already pass wallet ids to Core for hardware-wallet activity reads/writes/deletes. The backup problem is the typed decode boundary, not a need for platform-specific JSON patching.

`TransactionDetails` is not in the current Android/iOS activity backup envelopes. It still needs DB migration coverage because Core persists it and scopes it by `(walletId, txId)`.

## Requested change

Add Core-owned migration support at both Core-owned boundaries:

- migrate the activity DB on init, because Core owns the persisted activity layer.
- migrate Core-owned JSON slices embedded in Android/iOS VSS backup envelopes, so the apps can rewrite those backups with current entries.

The apps should not hand-edit Core model JSON or decode legacy Core models before migration.

Required helpers:

- migrate activity arrays from backup JSON into current `Activity` entries.
[...]

PR #117, opened 2026-06-26
refactor: always include walletId in current activity payloads

### LINKED ISSUE #112: refactor: move walletId to v2 activity models
## Problem

`walletId` was added to the current `v1` activity models. That changes the `v1` schema without changing the version.

This has not shipped to production mobile users yet, so we can still fix the model before Android and iOS integrate it as stable behavior.

## Requested change

Move `walletId` to the next activity model version.

- Keep the existing `LightningActivityV1` and `OnchainActivityV1` payloads unchanged.
- Add `LightningActivityV2` and `OnchainActivityV2` with `walletId`.
- Treat v1 activities as default-wallet activities when Core migrates or imports them.
- Store new wallet-scoped activity data as the latest activity model version.
- Keep lookup, list, delete, tags, seen state, and transaction details wallet-scoped without making v1 JSON invalid.

## API and search invariants

Moving `walletId` to v2 should not leave callers dealing with a mixed v1/v2 activity set.

- Public search/list/detail APIs should return the canonical/latest activity shape after migration, not both old and new versions of the same logical activity.
- Wallet-scoped APIs should not silently fall back to raw id or txid searches across wallet scopes.
- Activity identity should be consistently scoped by `(walletId, activityId)` once data is migrated.
- If Core keeps versioned enum variants internally, queries must still dedupe or normalize so callers do not see duplicate v1/v2 entries.
- Tests should cover a v1 and v2 activity with the same raw id and verify wallet-scoped lookup/list/search returns only the expected current record.

## Migration expectation

The storage and backup migration details belong in:
- #113
This issue should still avoid introducing a model shape that makes that migration impossible or ambiguous.

## Acceptance criteria

- `walletId` is not required on v1 activity payloads.
- New wallet-scoped activities serialize as a versioned model that includes `walletId`.
- Old v1 activity JSON without `walletId` decodes successfully.
- Wallet-scoped APIs treat imported/migrated v1 activities as default-wallet activities.
- Public queries do not return mixed or duplicate v1/v2 entries for the same logical activity.
- Tests cover old v1 decode, v2 hardware-wallet activity creation, and mixed v1/v2 search behavior.

coreyphillips:
## Summary

Resolves #112. `walletId` was added to the activity models, changing the v1 JSON schema. This keeps the models on the current canonical shape and proves the wallet-scoping invariants with tests, without blocking the backup/migration work in #113.

## Contract

- Current/exported activity payloads always include `walletId`, including the default `bitkit` wallet, so a fresh export is never indistinguishable from legacy data.
- Legacy JSON without `walletId` still decodes, defaulting to `bitkit` via the serde `default`.
- Normal query APIs return only the canonical shape. No `V1 | V2` variants are exposed to callers: one struct, `wallet_id` is a column with composite primary key `(wallet_id, id)`.
- Empty-`walletId` normalization and the broader backup/import/export plus DB migration are deferred to #113, where they belong.

Applies to `OnchainActivity`, `LightningActivity`, `ActivityTags`, `PreActivityMetadata`, and `TransactionDetails`.

## Tests

- Old v1 JSON without `walletId` decodes to the default wallet.
- Default-wallet and hardware-wallet records always serialize `walletId`.
- A v1 and v2 activity sharing one raw id stay distinct on wallet-scoped lookup, list, and search.

`cargo test modules::activity`: 178 passed, 0 failed.


ovitrif:
I like the ingenuity of this proposal at the persistence layer, but I think it missed half of the story: the issue we are aiming to resolve is also very much impacting the next task that comes after this:
- #113

So, the governing thought in `bitkit-core` should be how to enable ease of development in the consumers, which are and will always be only the Bitkit apps.

For now only Bitkit Android and iOS are the real consumers. Bitkit CLI might become a future target, but it should not drive this design today. So this means we shape this project with bindings and app integration concerns as core principles.

Therefore, I think versioning should still be added, but instead of versioning at the persistence level, which was indeed not the best requirement in the original issue, versioning should be implemented at the transport layer: `DTOs`, contracts between the Bitkit apps' native layer and Core's `FFI` layer.

Important consideration: the apps store this data modeled as backup/restore records, serialized and encrypted through `VSS` server-side. Apps should be able to read and decode old versions of this data, then provide wiring for migrating to the newer version, then wiring up the update to the new model, then continue working with the data by treating it only using the new version.

Ideally, `bitkit-core` integrates all these concerns so we can remove code duplication from the apps. This is shared logic, it has to work the same for both mobile platforms, and it has to work when changing from one platform to the other. For example: I backup a wallet on Android. I restore the wallet on iOS. Provided I do not do these in parallel, restore should work seamlessly. Core should own this.

This is all just context. We will move towards this ideal in the PR to close the next task (#113), but we should make sure whatever we design here allows that task to be implemented.

Final concern/requirement: these PRs would be better if there is an exercise done locally to see how the apps would integrate their changes. What I am entertaining: it would be best if there are local changes or even draft PRs for Android/iOS to prove/document the usage of the updated API models. This way, we can iterate towards a better solution and it will also help the client-side work to integrate in the consumers.

ovitrif (changes requested):
See long top-level comment describing the direction for the requested changes.

ovitrif:
One concrete contract I want to make explicit here: I do not think normal Core query APIs need to expose `V1 | V2` variants. After import/migration, they should return only the canonical current models. The versioning concern is specifically the backup/import/export `DTO` boundary.

At that boundary, I do not think absence of `walletId` should be treated as the current serialized shape. If this PR serializes default-wallet records without `walletId`, then a freshly exported default-wallet backup is indistinguishable from legacy data, and Android/iOS restore code keeps needing legacy handling indefinitely.

I think the safer contract is:

- legacy import/migration may accept missing or empty `walletId` and normalize it to the default wallet
- current/exported backup `DTOs` always include `walletId`, even when it is `bitkit`
- once restored/migrated, normal APIs and rewritten backups use only the current shape

Could we update the design/implementation toward that before merging, or at least make sure this PR does not block that contract for #113?

coreyphillips:
Re: transport-layer versioning and app integration

Agree with the framing. Core should own the backup/restore/migration so both apps share one implementation. I have scoped this PR (issue #112) deliberately narrow: get the activity model shape right so it does not block that work, nothing more. The DTO/transport versioning, the Core-owned migrate/import/export helpers, the DB migration, and the app-side integration exercise all belong to #113. Skipping the Android/iOS draft-PR exercise for now, since it is a lot of churn until bitkit-mono supports this more cleanly; we will prove the usage when #113 lands.

Re: always include walletId and the safer contract

Adopted the contract. Dropped `skip_serializing_if`, so current/exported records now always serialize `walletId` (including `bitkit`), and a freshly exported default-wallet backup is no longer indistinguishable from legacy data. Kept the serde default so legacy JSON missing `walletId` still decodes to `DEFAULT_WALLET_ID`. Normal query APIs already return only the canonical shape (single struct, `wallet_id` is a column with composite primary key `(wallet_id, id)`), so no `V1 | V2` variants are exposed to callers. The empty-`walletId` to default normalization is intentionally left for #113's migration helpers, since it is scoped to legacy migration input only (normal writes should still reject empty).

Tests now cover legacy v1 decode, current-shape serialization always carrying `walletId`, and a v1/v2 pair sharing one raw id staying wallet-scoped on lookup, list, and search. `cargo test modules::activity`: 178 passed, 0 failed. Branch is rebased onto current master.

ovitrif (approved):
Looks good to me now. This addresses my blocking concern: current/exported activity payloads carry `walletId`, legacy missing `walletId` still decodes as the default wallet, and normal APIs stay on the canonical current model.

#113 still owns the backup/import/export migration APIs and app-shaped backup tests.

coreyphillips:
Rebuilding the bindings now...

ovitrif:
Small non-blocking test note after approval: the mixed lookup/list/search test proves the default wallet does not see the hardware row, which is the important leak check.

It might be worth adding the positive side too, just to make the test read complete: assert that `Some(wallet_id)` + `bc1qv2hardware` returns exactly the hardware row, and optionally that `Some(DEFAULT_WALLET_ID)` + `bc1qv1default` returns exactly the default row.

ovitrif:
All good, thanks!

Please ensure we have a release after merge 🙏🏻

PR #120, opened 2026-07-02
Upgrade trezor-connect-rs to 0.4.0

coreyphillips:
## Summary

Upgrades trezor-connect-rs from 0.3.3 to 0.4.0 ([release notes](https://github.com/coreyphillips/trezor-connect-rs/releases/tag/v0.4.0)).

0.4.0 closes gaps against the JS upstream (@trezor/connect) and was verified end-to-end on a Trezor Safe 7 over USB:

- client-side validation before signing (dust limit, prev-tx presence and hash verification, address parsing/network checks)
- post-sign verification: the device-returned transaction is cross-checked against the request (counts, amounts, output scripts), and txid is now computed locally
- get_address hardening: silent pre-derive with expected-address check, multisig/chunkify forwarded, mac surfaced
- USB THP pairing credentials are stored and reused, so Safe devices over USB pair once instead of on every process run; busy handshakes retry with backoff

## Changes in this repo

- Bump both target-specific trezor-connect-rs requirements to 0.4.0
- The From impls for GetAddressParams, SignMessageParams, VerifyMessageParams and SignTxParams now use struct-update syntax (`..Default::default()`), covering the new optional fields (chunkify, expected address) and dropping the deprecated ignored `push` field, so future additive fields will not break the build
- Test fixture updated for the new SignedTxResponse.witnesses field

## Behavior note for app layers

0.4.0 no longer silently defaults unset input sequences to 0xFFFFFFFD (RBF); the field is omitted so firmware applies its 0xFFFFFFFF (final) default, matching @trezor/connect. bitkit-core passes the caller's sequence through unchanged, so apps that want RBF must set `sequence = 0xFFFFFFFD` explicitly when building inputs.

## Test plan

- cargo build clean, no deprecation warnings
- trezor module tests: 60 passed, 0 failed
- Full suite: only the pre-existing blocktank staging-API network tests fail (api.stag.blocktank.to unreachable), unrelated to this change


coreyphillips:
Once approved, I will bump the versions, bindings and create a PR containing the changes for the iOS & Android apps accordingly.

jvsena42 (approved):
Implemented on Android and iOS

PR #121, opened 2026-07-08
feat: wallet-scoped backup migration APIs

### LINKED ISSUE #113: fix: add wallet-scoped backup migrations

coreyphillips:
## Summary

Adds Core-owned backup migration for the wallet-scoped activity data, covering the backup-JSON boundary of #113. The activity DB migration half already landed; this PR handles the app backup envelopes.

### Background

Android and iOS embed Core-owned records (activities, tags, pre-activity metadata, transaction details) inside their own VSS backup envelopes and decode them directly into the generated binding types. Those records gained a required wallet id, so a backup written before the field existed no longer decodes: the platform decoder (kotlinx.serialization on Android, Codable on iOS) rejects the missing field before Core is ever consulted. The two platforms also serialize the `Activity` union in incompatible shapes, so no single serde type can read both:

| | Android (kotlinx) | iOS (Codable) |
|---|---|---|
| `Activity` union | `{"type":"...Onchain","v1":{...}}` | `{"onchain":{"_0":{...}}}` |
| `PaymentType` | `"SENT"` | `{"sent":{}}` |

### What this adds

Two Core-owned boundaries in `src/modules/activity/backup_migration.rs`, so the apps never hand-edit Core model JSON:

- **Format-tolerant normalizers** (`migrate_backup_*_json`) take the raw slice exactly as the app serialized it, inject the default wallet id wherever a record is missing it or has it empty, and return JSON in the same shape. They handle the Android, iOS and canonical `Activity`-union encodings plus flat records (tags, metadata, transaction details).
- **Canonical serialization** (`*_to_json` / `*_from_json`) is Core's own stable cross-platform JSON for these slices, with the `from_json` side defaulting a missing/empty wallet id so apps can move their backups onto one format over time.

### Migration rule

A missing or empty wallet id on legacy input becomes `DEFAULT_WALLET_ID` (`bitkit`); any existing non-default id such as `trezor:{hash}` is preserved. Normal writes still reject an empty wallet id via `normalize_wallet_id`; the empty to default relaxation applies to migration input only. Closed channels are not wallet-scoped, so their round-trip applies no normalization.

## Testing

- 11 new tests in `backup_migration.rs` covering Android/iOS/canonical activity shapes, empty to default normalization, non-default preservation, flat tag/metadata slices, non-array rejection, and canonical round-trips.
- `cargo test modules::activity` passes; clippy clean.

## Notes

- Bumps to 0.4.0 and regenerates Android, iOS and Python bindings.
- App-side wiring for bitkit-android and bitkit-ios to call these helpers before decoding is prepared and will follow in their own PRs once this releases.


piotr-iohk (approved):
LGTM

PR #123, opened 2026-07-10
fix(onchain): return canonical txid for already-known raw broadcasts

### LINKED ISSUE #122: Return txid for already-known raw broadcasts
# Return txid for already-known raw broadcasts

## Problem

`trezor_sign_tx_from_psbt` returns a valid serialized signed transaction but normally leaves `TrezorSignedTx.txid` unset. This matches the current contract: the txid is populated only when signing also pushes the transaction.

The native apps broadcast separately and retain the signed transaction after an ambiguous network failure. If Electrum accepted the first broadcast but its response was lost, retrying can return an “already known”, “already in mempool”, or “already in blockchain” error. The apps then cannot complete transfer bookkeeping because they do not have the txid.

Requiring `TrezorSignedTx.txid` at signing time is not valid and breaks normal hardware signing.

## Proposed change

Update `broadcast_raw_tx` in `src/modules/onchain/implementation.rs`:

1. Decode and deserialize the signed transaction as it already does.
2. Compute its canonical txid with `bitcoin::Transaction::compute_txid()`.
3. Return the computed txid after a successful Electrum broadcast.
4. Treat only explicit duplicate/already-accepted Electrum responses as success and return the computed txid.
5. Preserve existing errors for invalid transaction data, connectivity failures, task failures, and unrelated broadcast rejection.

Keep `TrezorSignedTx.txid` optional and preserve its existing contract. Do not implement Bitcoin transaction parsing independently in Swift or Kotlin.

## Error classification

Extract duplicate-response classification into a focused helper. Cover the Electrum messages currently handled by the apps:

- `already in block chain`
- `already in blockchain`
- `already in mempool`
- `already-in-block-chain`
- `already-in-mempool`
- `txn-already-known`
- `transaction already exists`

Classification must be case-insensitive and must not convert unrelated relay-policy or invalid-transaction errors into success.

## Tests

- A valid legacy transaction returns its canonical txid.
- A valid SegWit transaction returns `txid`, not `wtxid`.
- Each supported already-known response returns the locally computed txid.
- Connectivity and unrelated broadcast errors remain failures.
- Invalid hex and invalid serialized transactions retain their typed errors.

## Acceptance criteria

- Retrying an already-accepted raw transaction returns the same txid as the original broadcast.
- Native apps can complete Blocktank funding bookkeeping without relying on `TrezorSignedTx.txid`.
- No FFI API shape change is required.
- Rust tests, Clippy, and formatting pass.

## Consumer follow-up

After releasing Bitkit Core:

1. Update the Core dependency in `bitkit-ios` and `bitkit-android`.
2. Remove app-level already-known error parsing and signer-provided txid recovery.
3. Keep regression tests where Trezor signing returns `txid = nil` and broadcast retry still completes.

coreyphillips:
## Summary

Retrying an already-accepted raw broadcast previously could not recover the txid. `broadcast_raw_tx` now computes the transaction's canonical txid locally and returns it on success. When Electrum reports the transaction is already known (already in a block, in the mempool, or otherwise accepted), that response is treated as success and the same locally computed txid is returned.

This lets the native apps complete Blocktank funding bookkeeping after an ambiguous network failure without relying on `TrezorSignedTx.txid`, which is intentionally unset during pure hardware signing. The change is self-contained in the onchain module, has no dependency on trezor-connect-rs, and makes no FFI API shape change.

## Changes

- Add `decode_and_compute_txid`: decodes hex, validates the transaction, and computes the canonical txid (returns the txid, never the wtxid).
- Add `is_already_known_broadcast_error`: a focused, case-insensitive classifier for duplicate/already-accepted Electrum responses. It does not turn relay-policy or invalid-transaction rejections into success.
- Rewrite `broadcast_raw_tx` to return the locally computed txid on a fresh broadcast and on an already-known response. Connectivity failures, unrelated broadcast rejections, invalid hex, and invalid transaction data keep their existing typed `BroadcastError` variants.
- Bump to 0.4.1 and regenerate bindings (no generated source changes; FFI shape unchanged).

Handled already-known responses (case-insensitive): `already in block chain`, `already in blockchain`, `already in mempool`, `already-in-block-chain`, `already-in-mempool`, `txn-already-known`, `transaction already exists`.

## Tests

- Valid legacy transaction returns its canonical txid.
- Valid SegWit transaction returns the txid, not the wtxid.
- Each supported already-known response is classified as success.
- Unrelated relay-policy, invalid-transaction, and connectivity errors remain failures.
- Existing invalid-hex and invalid-transaction typed-error tests retained.

`cargo test`, `cargo clippy`, and `cargo fmt` pass.

## Related

- Closes #122
- Consumer follow-up (after release): bitkit-ios #621 and bitkit-android #1067 can drop the app-level already-known parsing and signer-provided txid recovery once they pin 0.4.1.


coreyphillips:
Release created for testing [here](https://github.com/synonymdev/bitkit-core/releases/tag/v0.4.1).

piotr-iohk (approved):
lgtm

PR #132, opened 2026-08-07
fix: stop writing data to stdout

coreyphillips:
Addresses the logging findings from the key generation security audit (prior ISSUES/01, reported again as F-02 and its adjacent "println hygiene" finding).

## Why

bitkit-core is a library that ships inside a wallet. `println!` and `eprintln!` bypass the `log` facade that the platforms route to their own sinks, so anything printed can be captured by a debugger, a redirected stdio stream, or a sysdiagnose bundle.

Two of these prints carried data that must never leave the process:

* `generate_mnemonic` printed the full BIP39 mnemonic, the root secret of the wallet, at the moment of generation. The audit rated this High rather than Critical only because no production caller exists today, but the UniFFI export is live and nothing marks it test-only.
* `create_and_store_order` printed the entire Blocktank order response, including invoice, addresses, and amounts, on every order creation.

The rest were debug leftovers, several of which printed the Bitcoin address under validation.

## What changed

All 21 print macros in non-test library code are gone: 15 deleted, 6 routed through `log::error!` / `log::warn!`. The converted sites log the error value only, never the input that produced it.

There are no behavior changes: every affected function already returned a `Result`, so the prints were purely additive tracing.

To prevent regression, the library crate root now carries:

```rust
#![deny(clippy::print_stdout, clippy::print_stderr, clippy::dbg_macro)]
```

This is enforced by the existing `cargo clippy` step rather than a new CI grep. It applies to the lib target only, so `example/main.rs` keeps its prints, and the three test modules that print carry a scoped `#![allow(clippy::print_stdout)]`.

## Testing

* `cargo clippy --all-targets` is clean of the new lints and of errors.
* `cargo fmt --check` passes.
* `cargo test modules::onchain` passes 66/66.
* Full suite: 474 passed / 11 failed. The 11 failures are the Blocktank staging API tests, confirmed failing identically on clean master, so they are pre-existing and network-dependent rather than regressions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -564,7 +564,7 @@ dependencies = [
 
 [[package]]
 name = "bitkitcore"
-version = "0.5.5"
+version = "0.5.6"
 dependencies = [
  "android_logger",
  "async-trait",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bitkitcore"
-version = "0.5.5"
+version = "0.5.6"
 edition = "2021"
 
 [lib]
--- a/Package.swift
+++ b/Package.swift
@@ -3,8 +3,8 @@
 
 import PackageDescription
 
-let tag = "v0.5.5"
-let checksum = "a1224eeee1bbaa9ecd2d09df9886b55bf825521670c22ec55bcc7853d5695489"
+let tag = "v0.5.6"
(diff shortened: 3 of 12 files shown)


coreyphillips:
Will make a new build once approved.

ovitrif (approved):
Approved. Library code no longer writes the mnemonic or Blocktank order payload to stdout, so those values cannot leak through redirected stdio.

ovitrif (approved):
This change looks good because production print macros are gone and the new clippy deny lints prevent them from returning.

PR #138, opened 2026-08-25
feat: add Passport hardware wallet support

ben-kaufman:
### Description

- Generalize the hardware-wallet catalog around vendor and transport metadata and add Foundation Passport over QR.
- Add generic multipart UR decoding plus `crypto-psbt` encoding, with Passport generic JSON account-export parsing and shared encoded-frame limits.
- Add signed PSBT combination, finalization, and verification for QR signer responses, including validation that signer-provided previous-output metadata matches the original PSBT.
- Keep the pending bitkit-core release on `0.5.8` under the internal Bitkit-app consumer policy, and regenerate the Swift, Kotlin/Android, and Python bindings and native artifacts from a clean build.

### Preview

N/A — Core/FFI change.

### QA Notes

- Passed a real Passport firmware-simulator round trip: account export, multipart `crypto-psbt`, firmware signing, signed-response decoding, and final transaction verification.
- `cargo test modules::hardware_wallet` — 1 passed.
- `cargo test modules::ur` — 14 passed.
- `cargo test modules::onchain::psbt` — 5 passed.
- `cargo clippy --all-targets` passed.
- Clean iOS binding/XCFramework generation, Swift package manifest validation, archive verification, and checksum validation passed.
- Android four-ABI build, native validation, Kotlin compilation, AAR build, and Maven Local publish passed.
- Python binding generation, checksum validation, and smoke import passed.
- Full suite: 493 passed and 26 ignored. The remaining 11 failures are existing live Blocktank staging tests whose staging API was unavailable.


src/modules/ur/decoder.rs
```diff
@@ -0,0 +1,406 @@
...
+        })
+    }
+
+    /// Accept one camera frame. Invalid or changed streams reset the decoder.
+    pub fn receive(&self, frame: String) -> Result<UrDecoderStatus, UrError> {
+        let mut state = self
+            .state
+            .lock()
+            .unwrap_or_else(std::sync::PoisonError::into_inner);
+        let result = state.receive(&frame);
```
ovitrif: Any `receive` error clears in-progress fountain state, including unreadable camera frames that fail bytewords checksums. Animated QR scanning produces those frames routinely, and the underlying UR decoder already leaves fragments in place when a part is invalid. Wiping progress on each miss prevents a noisy scan from accumulating parts, so Passport pairing and signed-PSBT return cannot complete under ordinary camera conditions. Please keep the current state on parse and checksum failures, and reset only when a successfully parsed part belongs to a different UR stream.

ovitrif (commented):
`UrDecoder.receive` wipes fountain progress on every invalid camera frame, including unreadable frames that fail bytewords checksums. Animated Passport QR scans produce those frames routinely, so pairing and signed-PSBT return cannot complete under ordinary camera conditions. Keep the current state on parse and checksum failures, and reset only when a successfully parsed part belongs to a different UR stream.

src/modules/ur/decoder.rs
ben-kaufman: Fixed in `55e904a8`. Parse and checksum failures now leave multipart progress intact. A successfully parsed fragment from a different stream replaces the incomplete stream while retaining that first fragment. I added regression tests for both cases.

src/modules/onchain/psbt.rs
```diff
@@ -0,0 +1,178 @@
...
+/// Combine and finalize a signed PSBT, then extract its broadcastable transaction.
+#[uniffi::export]
+pub fn finalize_psbt(
+    original_psbt: String,
+    signed_psbt: String,
+) -> Result<CompletedTransaction, PsbtCompletionError> {
+    let mut combined = parse_psbt(&original_psbt, "original")?;
+    let signed = parse_psbt(&signed_psbt, "signed")?;
+    let secp = Secp256k1::verification_only();
+    combined
```
coreyphillips, line 19: Could we validate signed PSBT input metadata against the original before combining it? For a legacy input, `combine` can accept a signer-supplied `witness_utxo` and discard the original `non_witness_utxo`. This lets verification succeed against the wrong previous output and return a transaction the network rejects.

src/modules/ur/decoder.rs
```diff
@@ -0,0 +1,500 @@
...
+#[uniffi::export]
+impl UrDecoder {
+    #[uniffi::constructor]
+    pub fn new() -> Arc<Self> {
+        Arc::new(Self {
+            state: Mutex::new(DecoderState::default()),
+        })
+    }
+
+    /// Adds one UR fragment and returns the current decoding status.
```
coreyphillips, line 34: Could we regenerate every wrapper and native artifact from the final head? A clean head build reports checksum `44279` for `UrDecoder.receive`, while the committed wrappers and binaries use `11477`.

src/modules/hardware_wallet/types.rs
```diff
@@ -0,0 +1,28 @@
...
+pub struct SupportedHardwareWallet {
+    pub vendor: HardwareWalletVendor,
+    /// Human-readable manufacturer name, e.g. "Foundation".
+    pub vendor_name: String,
+    /// Stable model identifier that applications can map to bundled assets.
+    pub model: String,
+    /// Full user-facing name.
+    pub display_name: String,
+    /// Transports over which the application can interact with this model.
+    pub transports: Vec<HardwareWalletTransport>,
```
coreyphillips, line 27: This changes the public `transports` type in a patch release. Could we preserve the old API or release this as `0.6.0` with migration notes?

src/modules/ur/encoding.rs
```diff
@@ -0,0 +1,121 @@
...
+    let fragment_count = cbor.len().div_ceil(max_fragment_length);
+    if fragment_count > MAX_FRAGMENT_COUNT {
+        return Err(UrError::TooLarge {
+            reason: format!(
+                "message requires {fragment_count} fragments; maximum is {MAX_FRAGMENT_COUNT}"
+            ),
+        });
+    }
+
+    if cbor.len() <= max_fragment_length {
```
coreyphillips, line 59: Could we apply the limit to the final encoded UR frame, or reject fragment sizes that exceed the decoder’s limit?

ben-kaufman: Fixed in 24d41d4f. The encoder now validates final single-part and multipart UR strings against the same frame limit used by the decoder. Added focused tests for oversized output in both paths.

src/modules/hardware_wallet/types.rs
ben-kaufman: Fixed in 24d41d4f. Bumped the pending release line to 0.6.0 and added a changelog migration note for SupportedHardwareWallet.transports moving to HardwareWalletTransport, including the new Qr case.

src/modules/onchain/psbt.rs
ben-kaufman: Fixed in 24d41d4f. finalize_psbt now resolves previous outputs from both PSBTs before combining and rejects conflicting signer metadata. Added a legacy P2PKH regression where the signer supplies a different witness_utxo with a valid signature.

src/modules/ur/decoder.rs
ben-kaufman: Fixed in 24d41d4f. Rebuilt all targets from a clean Cargo target and regenerated the Swift, Kotlin, Python, Android native libraries, and XCFramework artifacts from the final source. The generated wrappers and native artifacts now agree on the UrDecoder.receive checksum 44279.

src/modules/hardware_wallet/types.rs
ovitrif: shouldn't be needed ;) while it is an open-source library, the only current consumers are the bitkit-* apps repos.

PS. I have this in my global AGENTS.md:
```
- Treat the following repos under synonymdev org as libraries only consumed by the bitkit apps: bitkit-core, vss-rust-ffi, ldk-node.
```

ben-kaufman: Agreed. Since the Bitkit apps are the known exact-pinned consumers and neither currently uses this catalog field, I reverted the pending version to 0.5.8 and removed the migration-specific changelog note in 1afed9d7. The generic transport type itself stays as-is. This thread was already resolved, so I left it resolved.

PR #139, opened 2026-08-26
fix: add android r8 rules

### LINKED ISSUE #140: add android r8 consumer keep rules
Add Android consumer keep rules to the AAR so minified apps can load this library without adding dedicated keep rules for this library.

ovitrif:
Fixes #140

The Android AAR now ships targeted R8 consumer keep rules for the UniFFI/JNA FFI surface.

This is rebased on master after synonymdev/bitkit-core#138, at `0.5.9`.

### Description

- Packages `bindings/android/lib/consumer-rules.pro` into the AAR as `proguard.txt`.
- Keeps JNA `Structure` fields and constructors, `Callback` methods, and `UniffiLib` / `IntegrityCheckingUniffiLib` native method names in `com.synonym.bitkitcore`.
- Fails `bundleReleaseAar` when those consumer keep rules are missing or empty.
- Bumps the package to `0.5.9`.

### Preview

N/A — Android library packaging only.

### QA Notes

Local verification: `./gradlew :lib:validateConsumerKeepRules :lib:bundleReleaseAar` from `bindings/android`, then confirmed `proguard.txt` inside the release AAR.


ovitrif:
version bump won't work until this PR is merged and bindings are rebuilt:
- #138

ovitrif:
@ben-kaufman or @coreyphillips pls do a review even though it's merged, my AI was feeling too lucky today

bindings/android/lib/consumer-rules.pro
```diff
@@ -0,0 +1,31 @@
...
+# Native.register maps remaining native methods by exact C symbol name.
+-keepclasseswithmembers,allowshrinking,includedescriptorclasses class com.synonym.bitkitcore.UniffiLib {
+    native <methods>;
+}
+-keepclasseswithmembers,allowshrinking,includedescriptorclasses class com.synonym.bitkitcore.IntegrityCheckingUniffiLib {
+    native <methods>;
+}
+
+# @Structure.FieldOrder is read at runtime.
+-keepattributes RuntimeVisibleAnnotations
```
ben-kaufman, line 25: R8 full mode still removes every `@Structure.FieldOrder` annotation here. In a minified consumer, the `RustBufferStruct` fields survive but the DEX contains zero `FieldOrder` runtime annotations, so JNA gets an empty field order and fails when it lays out the struct. Could we keep both the `FieldOrder` annotation type and the annotated bitkit-core classes, with shrinking, optimization, and obfuscation allowed, and add a minified-consumer assertion for this?

ben-kaufman:
The 0.5.9 release symbols no longer match the published Android AAR. The manual run published the immutable Maven package, then the release run rebuilt and replaced `native-debug-symbols.zip` before its Maven publish failed with 409. All four ABI build IDs differ. Could we restore the symbols from run `32992320812` and update the workflow so the AAR and symbols always come from one successful build?

bindings/android/lib/consumer-rules.pro
ovitrif: Opened synonymdev/bitkit-core#145 with the FieldOrder keep rules and a minified-consumer assertion.

Fixed in 413e1e4

ovitrif:
Addressed the review concerns in:
- #145

PR #141, opened 2026-08-26
feat: add wallet-scoped activity tag reads (#136)

### LINKED ISSUE #136: feat: add wallet-scoped activity tag reads
Found while wiring hardware-wallet tag backup on Android (synonymdev/bitkit-android#1163, follow-up to #113). Verified against `v0.5.5`.

## Problem

`get_all_activities_tags()` takes no wallet id, and the only scoped getter is `get_tags(wallet_id, activity_id)` for a single activity. There is no way to ask for "tags for wallet X", so apps fetch everything and filter in app code:

```kotlin
coreService.activity.getAllActivitiesTags()
    .filter { it.walletId == WalletScope.default }
```

That single line is the app's entire backup-scoping policy. It has already been added and removed once in Android, each time silently changing which wallet scopes leave the device — which matters, because these records go to a remote backup server.

## Requested change

Expose `get_activities_tags(wallet_id: Option<String>)`, where `None` returns every scope, so scoping becomes an explicit argument visible in review rather than an app-side filter. The same argument applies to `get_all_pre_activity_metadata()`.

## Acceptance criteria

- Tags can be read for a single wallet scope and for all scopes, without app-side filtering.
- The existing unscoped behaviour remains available.
- Tests cover scoped and unscoped reads, including a hardware `trezor:{hash}` scope.

coreyphillips:
Closes #136

Add `get_activities_tags(wallet_id)` and `get_pre_activity_metadata_list(wallet_id)`, wallet-scoped reads of the tag-backup tables where `None` returns every scope.

The two tag-backup readers, `get_all_activities_tags()` and `get_all_pre_activity_metadata()`, take no wallet id, and the only scoped tag getter is `get_tags(wallet_id, activity_id)` for a single activity. An app that wants one wallet's tags has to fetch every scope and filter client-side, which makes a one-line app-side `.filter { it.walletId == ... }` the whole backup-scoping policy for records that go to a remote backup server. Dropping that line is invisible in review and silently widens what leaves the device.

## What changed

- `ActivityDb::get_activities_tags(Option<&str>)` in `src/modules/activity/implementation.rs` builds the same tag query with an optional `WHERE wallet_id = ?1`; `get_all_activities_tags()` now just calls it with `None`.
- `ActivityDb::get_pre_activity_metadata_list(Option<&str>)` does the same for `pre_activity_metadata`; `get_all_pre_activity_metadata()` delegates to it with `None`.
- Both scoped readers run the wallet id through the existing `normalize_wallet_id`, so a blank or whitespace-only scope is a typed error rather than a silent full-table read.
- New UniFFI exports `get_activities_tags(wallet_id: Option<String>)` and `get_pre_activity_metadata_list(wallet_id: Option<String>)` in `src/lib.rs`. The existing unscoped exports are untouched, so no caller breaks.
- README, module README and CHANGELOG document the new calls.

## How to test

- `cargo test modules::activity`, 193 pass, including the four new tests.
- `test_get_activities_tags_wallet_scoped` and `test_get_pre_activity_metadata_list_wallet_scoped` each store one record under the default scope and one under a `trezor:{64-hex}` scope, then assert each scope returns only its own record, that `None` matches the legacy unscoped getter, and that an unknown scope returns empty.
- `test_get_activities_tags_rejects_blank_wallet_id` and `test_get_pre_activity_metadata_list_rejects_blank_wallet_id` cover the whitespace-scope error.
- `cargo clippy --all-targets` and `cargo fmt --check` are clean for the touched code.

## Notes

Naming: the issue asks for `get_activities_tags`, which I used verbatim. The metadata equivalent could not drop `all_` the same way because `get_pre_activity_metadata(wallet_id, search_key, search_by_address)` already exists and UniFFI function names are global, hence `get_pre_activity_metadata_list`. I added new functions rather than changing the existing signatures, so bitkit-android and bitkit-ios keep compiling; the acceptance criterion that unscoped behaviour stays available is met by both the old calls and by passing `None`. Generated bindings under `bindings/` are not regenerated here. This repo does that in separate `chore: regenerate bindings` commits at release time via `build.sh -r`, and it needs the iOS/Android cross-compile toolchain, so the new functions are not yet visible to Swift/Kotlin/Python until that step runs. `cargo test` (full suite) has 11 pre-existing failures, all in `modules::blocktank`, from `api.stag.blocktank.to` being unreachable in this environment. They are unrelated to this change; the other 497 pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,100 @@
 
 - Add `get_activities_tags(wallet_id: Option<String>)` and `get_pre_activity_metadata_list(wallet_id: Option<String>)`, wallet-scoped reads of the two tag-backup tables where `None` returns every scope. Apps backing up a single wallet scope (for example a `trezor:{hash}` hardware wallet) can now name that scope in the call instead of fetching every scope and filtering client-side, so which records leave the device is visible in the FFI call. The unscoped `get_all_activities_tags()` and `get_all_pre_activity_metadata()` are unchanged and now delegate to the scoped versions.
 
-- The Android AAR now ships targeted R8 consumer keep rules for the UniFFI/JNA FFI surface, so consuming apps can enable R8 full mode without extra keep rules for this library.
+## 0.5.11 - 2026-08-27
+
+- Add keep consumer rules for JNA types UniFFI needs under R8.
+
+## 0.5.10 - 2026-08-27
+
+- Watch-only watcher events now include the synchronized next unused external address and full derivation path.
+
+## 0.5.9 - 2026-08-26
+
+- Ship the initial Android R8 consumer keep rules for the UniFFI/JNA FFI surface.
+
+## 0.5.8 - 2026-08-26
+
+- Preserve UR decoder scan progress across invalid frames.
+- Reject signed PSBT previous-output metadata that differs from the original transaction.
+- Enforce encoded UR frame-size bounds.
+
+## 0.5.7 - 2026-08-24
+
 - Add a generic hardware-wallet catalog with Foundation Passport support, multipart UR QR encoding and decoding, Passport single-signature account export parsing, and signed PSBT finalization across the UniFFI bindings.
+
+## 0.5.6 - 2026-08-14
+
+- Stop writing secrets and user data to stdout. Library logs go through the `log` facade.
+
+## 0.5.5 - 2026-08-03
+
+- Add `serialized_extended_pubkey` for canonical 78-byte BIP32 xpub/tpub payloads, exposed as Swift `Data` and Kotlin `ByteArray`.
+
+## 0.5.4 - 2026-07-30
+
+- Persist the submarine-swap refund destination so `boltzGetSwap` and `boltzListSwaps` return it after completion.
+
+## 0.5.3 - 2026-07-28
+
+- Validate Boltz create responses, claim and refund addresses, and fee rates before swapping.
+- Keep a settled reverse swap pending until its claim txid is recorded locally, and retry those claims on the updates stream.
+
+## 0.5.2 - 2026-07-20
+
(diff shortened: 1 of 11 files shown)


coreyphillips:
Once approved I'll bump the version and rebuild the bindings.

src/lib.rs
```diff
@@ -752,6 +752,19 @@ pub fn get_all_activities_tags() -> Result<Vec<ActivityTags>, ActivityError> {
     db.get_all_activities_tags()
 }
 
+/// Activity tags for a single wallet scope, or every scope when `wallet_id` is `None`.
+#[uniffi::export]
+pub fn get_activities_tags(wallet_id: Option<String>) -> Result<Vec<ActivityTags>, ActivityError> {
```
ovitrif: I can see that the new database methods are covered, but no test invokes the exported `get_activities_tags` or `get_pre_activity_metadata_list` functions in `src/lib.rs`; the added tests call `ActivityDB` directly. If either wrapper dropped `wallet_id` or delegated to its unscoped getter, every added test would still pass while native clients could receive records from every wallet scope. Could we add one exported-entry-point test that initializes two wallet scopes and asserts both functions preserve `Some("trezor:...")` and `None` semantics?

coreyphillips: Good call, added in 6a9769d. `test_exported_scoped_getters_preserve_wallet_id` in `src/lib.rs` runs `init_db`, seeds two wallet scopes through the exported entry points, and asserts both `get_activities_tags` and `get_pre_activity_metadata_list` preserve `Some("trezor:...")` and `None` semantics, plus that an unknown scope returns empty.

ovitrif (approved):
utACK

ovitrif:
> Once approved I'll bump the version and rebuild the bindings.

Can you please rebase /bump version on top of #145 changes if @ben-kaufman can confirm it's ready for merge? so then we save a bit of bindings regen time 🙏🏻

ben-kaufman:
Yes looks ready

ovitrif:
Thanks @ben-kaufman 🙏🏻 , so now it should be easier, just merge with master and bump to 0.5.12 + re-bindgen cc @coreyphillips 🫡

PR #142, opened 2026-08-26
fix: drop legacy unique indexes on activity id that block wallet-scoped rows (#137)

### LINKED ISSUE #137: fix: drop legacy unique indexes on activity id that block wallet-scoped rows
Databases created before activity data became wallet-scoped still carry two legacy unique indexes that the wallet-scoping change never dropped:

```sql
CREATE UNIQUE INDEX idx_onchain_id   ON onchain_activity(id)
CREATE UNIQUE INDEX idx_lightning_id ON lightning_activity(id)
```

Both tables are now keyed by `PRIMARY KEY (wallet_id, id)`, so these indexes contradict the schema: they make an activity id globally unique **across wallet scopes**, when the whole point of the change was that the same id can exist once per wallet.

### Impact

On-chain activity ids are the txid, so a single transaction that is visible to two wallet scopes cannot be stored. The realistic case is a user paying their own hardware wallet: the transaction lands under `bitkit` first, and the watcher's upsert for `trezor:{hash}` is then rejected.

Because the snapshot is written in one transaction, the collision fails the **entire** write, not just the colliding row — so the hardware wallet persists *zero* activities and shows an empty list forever, retrying and failing on every watcher poll:

```
Failed to persist HW snapshot for 'trezor:84e3cf34…':
BitkitCore.ActivityError.InsertError(errorDetails: "Failed to upsert onchain_activity:
UNIQUE constraint failed: onchain_activity.id")
```

Only databases that predate wallet scoping are affected, so this is invisible in clean-slate testing and reproduces only on upgraded installs. Both iOS and Android are affected, since the schema is core's.

`idx_lightning_id` has not been observed failing in practice — watch-only hardware wallets have no Lightning — but it is the same defect and should go with it.

### Root cause

Both indexes were created by older core versions, e.g. `0.1.40` at `src/modules/activity/implementation.rs:126` and `:129`:

```rust
"CREATE UNIQUE INDEX IF NOT EXISTS idx_onchain_id ON onchain_activity(id)",
"CREATE UNIQUE INDEX IF NOT EXISTS idx_lightning_id ON lightning_activity(id)",
```

By `0.5.3` they are gone from the source — that version creates no unique indexes at all — but nothing removes them from existing databases. There is no `DROP INDEX` statement anywhere in the crate, and `CREATE ... IF NOT EXISTS` obviously cannot undo an index that already exists. The exact release that removed the statements was not pinned down; the boundary is somewhere between `0.1.40` and `0.5.3`.

### Reproduction

Real-world, on a wallet whose database predates wallet scoping:

1. Pair a hardware wallet and let the watcher sync — its activities appear.
2. Send any on-chain amount from the Bitkit wallet to an address of that hardware wallet.
3. Wait for the watcher to report. The hardware wallet's activity list is now empty, and the log shows `UNIQUE constraint failed: onchain_activity.id` on every poll.

Minimal, against any current database:

```sql
-- reproduce the legacy state
CREATE UNIQUE INDEX IF NOT EXISTS idx_onchain_id ON onchain_activity(id);
```

then upsert the same activity id under two different `wallet_id` values. The second fails, though the composite primary key permits it.

Confirmed on a live regtest database: `57` activities under the standard-wallet scope, `0` under the hidden-wallet scope that hit the collision, with both stale indexes present in `sqlite_master`.

### Suggested fix

Drop both indexes in the migration that introduces the composite primary key, and for databases already past that point, unconditionally on init:

```sql
DROP INDEX IF EXISTS idx_onchain_id;
DROP INDEX IF EXISTS idx_lightning_id;
```

It may be worth auditing for other indexes removed from source without a corresponding drop, since the same oversight pattern would apply.

### Suggested regression tests

1. **Behavioural, the one that matters.** Upsert the same activity id under two different wallet ids and assert both rows persist and are readable by scope. This fails today on a legacy database and passes on a fresh one, so it guards the invariant regardless of how the schema got there.
2. **Migration.** Open a database, create `idx_onchain_id` / `idx_lightning_id` by hand to simulate the legacy state, run init/migration, then assert `sqlite_master` contains neither — and that a two-scope upsert then succeeds.
3. **Schema drift guard.** Assert that after init, `onchain_activity` and `lightning_activity` carry no unique index other than the implicit primary-key one. That catches any future reintroduction, and would have caught this.

### References

- App-side context: synonymdev/bitkit-ios#675
- Wallet-scoped activity work: #102, #113

coreyphillips:
Closes #137

Drop the legacy idx_onchain_id and idx_lightning_id unique indexes on activity DB init, so an activity id can exist once per wallet scope as the composite primary key intends.

Databases predating wallet-scoped activity data can still carry `idx_onchain_id` and `idx_lightning_id`, unique indexes over the activity id alone. They contradict the `PRIMARY KEY (wallet_id, id)` both tables now use: an on-chain id is the txid, so a transaction visible to two wallet scopes (paying your own hardware wallet) cannot be stored twice. The watcher writes its snapshot in one transaction, so the collision rejects the entire snapshot, not just the colliding row, and the hardware wallet shows an empty activity list that never recovers. The statements that created the indexes were removed from the crate by the wallet-scoping commit (cc1ff06), but nothing ever dropped them from existing databases.

## What changed

- `ActivityDB::initialize` now runs a new `LEGACY_INDEX_DROP_STATEMENTS` step (`DROP INDEX IF EXISTS idx_onchain_id` / `idx_lightning_id`) unconditionally, after the table migrations and before the index creation loop, in `src/modules/activity/implementation.rs`
- Three regression tests in `src/modules/activity/tests.rs`: `test_init_drops_legacy_activity_id_unique_indexes` (legacy indexes hand-created on a current-schema database, then a two-scope snapshot upsert), `test_legacy_schema_migration_drops_activity_id_unique_indexes` (pre-wallet-scoped schema plus both indexes), and `test_activity_tables_have_no_extra_unique_indexes` (drift guard)
- Test helper `extra_unique_index_names` reads `PRAGMA index_list` and returns unique indexes whose origin is not `pk`, so the drift guard fails on any future non-primary-key unique index over `activities`, `onchain_activity`, or `lightning_activity`

## How to test

- `cargo test --lib modules::activity`: 192 passed, 0 failed.
- To confirm the tests are not vacuous, I replaced the drop loop with `LEGACY_INDEX_DROP_STATEMENTS.iter().take(0)` and reran: `test_init_drops_legacy_activity_id_unique_indexes` fails on the leftover index, and restoring the loop makes it pass. That test is the one that covers the reported failure.
- `cargo test` (full suite): 496 passed, 11 failed. All 11 failures are `modules::blocktank` tests hitting `api.stag.blocktank.to`, which is unreachable from this sandbox; they are unrelated to this change.
- `rustfmt --check` is clean on both touched files. `cargo fmt --check` reports a pre-existing diff in `src/modules/activity/backup_migration.rs`, which this change does not touch (reproduced against `HEAD` before the change).
- `cargo clippy --all-targets` produces no errors and no new warnings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Add `get_activities_tags(wallet_id: Option<String>)` and `get_pre_activity_metadata_list(wallet_id: Option<String>)`, wallet-scoped reads of the two tag-backup tables where `None` returns every scope. Apps backing up a single wallet scope (for example a `trezor:{hash}` hardware wallet) can now name that scope in the call instead of fetching every scope and filtering client-side, so which records leave the device is visible in the FFI call. The unscoped `get_all_activities_tags()` and `get_all_pre_activity_metadata()` are unchanged and now delegate to the scoped versions.
+- Drop the legacy `idx_onchain_id` and `idx_lightning_id` unique indexes on activity init. Older releases created them when an activity id was globally unique, which contradicts the `PRIMARY KEY (wallet_id, id)` the activity tables now use and blocked storing the same on-chain id once per wallet scope, for example a transaction visible to both the Bitkit wallet and a watched hardware wallet. `CREATE ... IF NOT EXISTS` never removed them from databases that already had them, so they are dropped unconditionally.
 
 ## 0.5.11 - 2026-08-27
 


ovitrif:
Would this warrant changes in the app(s)? @coreyphillips just asking

coreyphillips:
No app logic changes should be needed. This is handled by Core during the existing `initDb` call. Both apps will only need to bump to the Core release containing this fix.

ovitrif (approved):
utACK

PR #145, opened 2026-08-26
fix: android consumer-rules

### LINKED ISSUE #140: add android r8 consumer keep rules

ovitrif:
Fixes #140

Follow-up to synonymdev/bitkit-core#139.

The Android AAR ships consumer keep rules for UniFFI/JNA so a minified app can resolve JNI names and lay out JNA structs.

Library publish packages `proguard.txt` in the AAR. Minified-app R8 belongs in the consuming app.

### Description

- Keeps `com.sun.jna.*` so `libjnidispatch` can resolve JNI names on `Native`, `Structure`, and `CallbackReference`.
- Keeps classes that extend `com.sun.jna.*`, including `com.sun.jna.ptr` ByReference types used as UniFFI callback out-params.
- Keeps `com.sun.jna.Structure$FieldOrder` and annotated `com.synonym.bitkitcore` struct classes, fields, and constructors.
- Fails `bundleReleaseAar` when the AAR is missing a non-empty `proguard.txt`.
- Fixes changelog by backfilling entries from releases.

### Out of scope

- Running R8 in this repo.

### Preview

N/A — Android consumer keep rules only.

### QA Notes

- Validated by running a minified Bitkit `devRelease` on Pixel_10_Pro_XL (16KB) with this artifact served via mavenLocal: terms, skip intro, new wallet, home with savings and spending at ₿0. No JNI/JNA crash.
- That build dropped the app keep for `com.synonym.bitkitcore.**` and the app-wide FieldOrder, Structure, and Callback keeps.
- Release AAR contains a non-empty `proguard.txt` with those consumer keep rules.

### Release

- [Release 0.5.11](https://github.com/synonymdev/bitkit-core/releases/tag/v0.5.11)


bindings/android/lib/consumer-rules.pro
```diff
@@ -21,8 +21,14 @@
     native <methods>;
 }
 
-# @Structure.FieldOrder is read at runtime.
+# JNA reads Structure.FieldOrder at runtime. R8 full mode strips that
+# annotation unless the annotation type and annotated classes are kept.
 -keepattributes RuntimeVisibleAnnotations
+-keep,allowshrinking,allowoptimization class com.sun.jna.Structure$FieldOrder
```
ben-kaufman, line 27: Keeping `FieldOrder` is necessary, but minified consumers still obfuscate JNA's JNI-reflected core. With these rules plus AGP's defaults, R8 renamed `Structure` and `CallbackReference` and removed `Native.fromNative`, `nativeType`, and `toNative`. `libjnidispatch.so` looks up those exact symbols during `Native.initIDs`, so the app can fail before struct layout. JNA's Android guidance requires `-keep class com.sun.jna.* { *; }`, and adding that rule preserved the symbols in the same build. Could we ship that core rule and exercise JNA initialization in the minified check?

bindings/android/lib/build.gradle.kts
```diff
@@ -174,8 +176,140 @@ val validateConsumerKeepRules by tasks.registering {
...
+            zip.getInputStream(entry).use { input ->
+                jnaJar.outputStream().use { input.copyTo(it) }
+            }
+        }
+
+        val seedRules = work.resolve("consumer-seed.pro")
+        seedRules.writeText(
+            """
+            -keep class com.synonym.bitkitcore.UniffiLib { *; }
+            -keep class com.synonym.bitkitcore.RustBufferStruct { *; }
```
ben-kaufman: Keeping `RustBufferStruct` directly here makes this test pass without the annotated-class consumer rule. I removed only the `@FieldOrder class ...` rule and reran the same R8 invocation. `RustBufferStruct` was byte-identical and retained the annotation because this seed itself satisfies R8 full mode's kept-endpoint requirement, while other live structures lost their annotations. Could we seed a consumer entry point and inspect the mapped output instead, so the check fails if either endpoint rule is removed?

bindings/android/lib/build.gradle.kts
```diff
@@ -174,8 +176,140 @@ val validateConsumerKeepRules by tasks.registering {
...
+                add("--pg-conf")
+                add(seedRules.absolutePath)
+                add(programJar.absolutePath)
+                add(jnaJar.absolutePath)
+            }
+        }
+
+        val minifiedJar = r8Out.resolve("minified.jar")
+        val javap = ProcessBuilder(
+            "javap",
```
ben-kaufman: `ProcessBuilder("javap", ...)` assumes `javap` is on `PATH`, but Gradle can run from a JDK configured only through `JAVA_HOME` or `org.gradle.java.home`. I reproduced this with a valid Java 17 JDK. Compilation succeeded, then validation failed with `Cannot run program "javap": No such file or directory`. Since release bundling and publishing depend on this task, could we resolve `javap` from Gradle's JVM/toolchain or inspect the classfile in-process?

bindings/android/lib/consumer-rules.pro
ovitrif: Shipped `-keep class com.sun.jna.* { *; }`.

Dropped the Gradle R8/javap job. Library publish packages `proguard.txt`. The minified-app proof is synonymdev/bitkit-android#1184 (after integrating latest version and removing the app-side keep's).

Fixed in c9b0855

bindings/android/lib/build.gradle.kts
ovitrif: Dropped the fake consumer R8 job. A synthetic seed does not match AGP, and the minified-app proof is synonymdev/bitkit-android#1184, after it's updated to include the latest libs versions once the PRs land and their releases are done. Meanwhile validation was done by building the PR locally while serving it with the updated bitkit-core via mavenLocal.

Skipped in c9b0855

bindings/android/lib/build.gradle.kts
ovitrif: Dropped the `javap` task. `bundleReleaseAar` only checks that `proguard.txt` is packaged.

Skipped in c9b0855

bindings/android/lib/consumer-rules.pro
```diff
@@ -21,8 +21,18 @@
...
+-keep,allowshrinking,allowoptimization,allowobfuscation @com.sun.jna.Structure$FieldOrder class com.synonym.bitkitcore.** {
+    <fields>;
+    <init>(...);
+}
+
+# libjnidispatch looks up Native/Structure/CallbackReference members by JNI name.
+# See https://github.com/java-native-access/jna/blob/master/www/FrequentlyAskedQuestions.md#jna-on-android
+-keep class com.sun.jna.* { *; }
```
ben-kaufman: This only keeps classes directly in `com.sun.jna`, so R8 still strips the public no-arg constructors from `com.sun.jna.ptr.IntByReference` and `ByteByReference`. `UniffiLib` always registers the Trezor callback vtable, and JNA reflectively instantiates these callback parameter types during registration, so even a minified app whose first call is `getDefaultWalletId()` crashes in `UniffiLib.<clinit>`. Could we include the other two JNA Android FAQ rules here and add a consumer-style minification check without app-side JNA or Bitkit keeps?

ovitrif: Shipped the other two JNA Android FAQ rules so `com.sun.jna.ptr.IntByReference` and `ByteByReference` keep their public no-arg constructors for `UniffiLib` callback-vtable registration:

```
-keep class com.sun.jna.* { *; }
-keep class * extends com.sun.jna.* { *; }
-keepclassmembers class * extends com.sun.jna.* { public *; }
```

Library publish still packages `proguard.txt`. Minified-app R8 stays in the consuming app.

Skipped bindings regen: The last change is only `consumer-rules.pro` (packaged as AAR `proguard.txt`). UniFFI did not change, so iOS/Android/Python bindings do not need another regen.

Fixed in a29de4c

ovitrif:
@ben-kaufman this is ready for another pass and was tested locally in bitkit-android. I handled the fixes a little differently from the previous review suggestions by applying the JNA Android FAQ guidance linked from the updated consumer-rules file.

The rebased fix is in 7fdf4f2; the final binding regeneration was published from 844fa90 as [`v0.5.11`](https://github.com/synonymdev/bitkit-core/releases/tag/v0.5.11).

bindings/android/gradle.properties
```diff
@@ -3,4 +3,4 @@ android.useAndroidX=true
 android.enableJetifier=true
 kotlin.code.style=official
 group=com.synonym
-version=0.5.9
+version=0.5.10
```
ben-kaufman: The four regenerated 0.5.10 JNI libraries identify NDK `r28b` build `13356709`, while the release workflow pins `r28c` and checks `28.2.13676358`. Direct AAR builds therefore package different binaries from the release path. Could we regenerate these libraries with r28c and validate the NDK note during packaging?

CHANGELOG.md
```diff
@@ -2,12 +2,95 @@
...
 ## Unreleased
 
-- The Android AAR now ships targeted R8 consumer keep rules for the UniFFI/JNA FFI surface, so consuming apps can enable R8 full mode without extra keep rules for this library.
+## 0.5.10 - 2026-08-26
+
+- Add keep consumer rules for JNA types UniFFI needs under R8.
+
+## 0.5.9 - 2026-08-26
+
+- Ship targeted R8 consumer keep rules for the UniFFI/JNA FFI surface, so consuming apps can enable R8 full mode without extra keep rules for this library.
```
ben-kaufman: `v0.5.9` did not support R8 full mode without extra JNA keeps. Its tagged rules only kept generated UniFFI members and the annotation attribute, and this PR adds the missing `FieldOrder`, JNA core, and FAQ rules. Could we describe 0.5.9 as the initial consumer-rule set and reserve the no-extra-rules claim for 0.5.10?

CHANGELOG.md
```diff
@@ -2,12 +2,95 @@
 
 ## Unreleased
 
-- The Android AAR now ships targeted R8 consumer keep rules for the UniFFI/JNA FFI surface, so consuming apps can enable R8 full mode without extra keep rules for this library.
+## 0.5.10 - 2026-08-26
```
ben-kaufman: `0.5.10` is filed as released on 2026-08-26, but no `v0.5.10` tag exists and the release content continued changing in commits dated 2026-08-27. Could we keep this under `Unreleased` until the tag is cut, then use the actual release date?

CHANGELOG.md
```diff
@@ -2,12 +2,95 @@
...
+
+- Add keep consumer rules for JNA types UniFFI needs under R8.
+
+## 0.5.9 - 2026-08-26
+
+- Ship targeted R8 consumer keep rules for the UniFFI/JNA FFI surface, so consuming apps can enable R8 full mode without extra keep rules for this library.
+
+## 0.5.8 - 2026-08-26
+
+- Preserve UR decoder scan progress across invalid frames.
```
ben-kaufman: `v0.5.8` also shipped `24d41d4`, which rejects signed PSBT previous-output metadata that differs from the original and adds encoded UR frame-size bounds. Could we include these user-visible validation changes in this entry?

bindings/android/gradle.properties
ovitrif: I audited this across Bitkit's owned native dependencies. This is a broader cross-repository toolchain-alignment concern rather than a regression specific to this R8 PR:

- Bitkit Android's stable/internal release paths use r28b / `28.1.13356709`, and the binaries in this PR were generated with that same NDK.
- Published Bitkit Core `0.5.9` uses r28c.
- Published LDK Node, VSS client, and Paykit artifacts actually use r27d: their workflows request r28c but leave the runner's r27d `ANDROID_NDK_HOME` conflicting with `ANDROID_NDK_ROOT`.
- All inspected owned Rust libraries remain 16 KB aligned (`PT_LOAD` `0x4000`) and do not depend on `libc++_shared.so`, so the audit found no current artifact failure.

The coordinated fix is tracked for Bitkit `2.6.0` in synonymdev/bitkit-android#1190. The Bitkit Core work is synonymdev/bitkit-core#148; the epic links the LDK, VSS, Paykit, and final Android integration subissues as well.

Keeping that cross-repo alignment work out of this consumer-rules PR and closing this thread.

CHANGELOG.md
ovitrif: Updated the 0.5.9 entry to describe it as the initial consumer-rule set, and moved this PR’s additional JNA rules under Unreleased. Fixed in 7ca7edb.

CHANGELOG.md
ovitrif: `v0.5.10` was released from a separate master commit and is not part of this PR, so I removed that section instead. This PR’s JNA rules remain under Unreleased. Fixed in 303a066.

CHANGELOG.md
ovitrif: Added the PSBT previous-output validation and encoded UR frame-size bounds in 7ca7edb. This section was backfilled from the GitHub release descriptions, which exposed that the `v0.5.8` description itself repeated Passport from `v0.5.7` and omitted these changes, so I corrected the GitHub release description as well: https://github.com/synonymdev/bitkit-core/releases/tag/v0.5.8.

ben-kaufman (approved):

PR #147, opened 2026-08-27
feat: expose watcher receive address

ben-kaufman:
### Description

- Include the synchronized wallet's next unused external receive address in every `TransactionsChanged` watcher event.
- Build the full derivation path from the resolved account path and BDK's `AddressIndex::LastUnused`, while preserving the existing watcher error event on derivation failure.
- Cover initial state, external address advancement, change-address isolation, derivation paths, the derivation-error path, and the existing activity/transaction payload.
- Bump Core to 0.5.10 and regenerate the iOS, Android, and Python bindings and native artifacts.

### Preview

N/A — generic Core API change.

### QA Notes

- `cargo test modules::onchain::listener::tests` (3 passed)
- `cargo clippy --lib --no-deps` (passed; existing repository warnings only)
- Targeted `rustfmt --check` and `git diff --check` (passed)
- `./build_ios.sh` (passed)
- `./build_android.sh` (passed, including Maven Local AAR publication)
- `./build_python.sh` (passed; generated package import smoke test passed)
- Full `cargo test`: 495 passed, 26 ignored; 11 existing live Blocktank staging tests failed because requests to `api.stag.blocktank.to` could not be sent.

### DIFF
--- a/src/modules/onchain/listener.rs
+++ b/src/modules/onchain/listener.rs
@@ -18,7 +18,8 @@ use super::implementation::{
     watch_only_activity_from_detail,
 };
 use super::types::{
-    AccountType, Network as OnchainNetwork, WalletBalance, WatcherEvent, DEFAULT_GAP_LIMIT,
+    AccountType, AddressInfo, Network as OnchainNetwork, WalletBalance, WatcherEvent,
+    DEFAULT_GAP_LIMIT,
 };
 use bdk::bitcoin::Network as BdkNetwork;
 
@@ -343,6 +344,7 @@ fn build_tx_changed_event(
     account_type: AccountType,
     wallet_id: &str,
     wallet_network: BdkNetwork,
+    base_path: &str,
 ) -> WatcherEvent {
     let bdk_balance = match wallet.get_balance() {
         Ok(b) => b,
@@ -382,6 +384,14 @@ fn build_tx_changed_event(
     }
 
     let tx_count = u32::try_from(activities.len()).unwrap_or(u32::MAX);
+    let next_unused_external_address = match next_unused_external_address(wallet, base_path) {
+        Ok(address) => address,
+        Err(error) => {
+            return WatcherEvent::Error {
+                message: error.to_string(),
+            };
+        }
+    };
 
     WatcherEvent::TransactionsChanged {
         activities,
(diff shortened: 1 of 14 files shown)


src/modules/onchain/types.rs
```diff
@@ -554,13 +554,16 @@ pub enum WatcherEvent {
...
+    /// `next_unused_external_address` is the synchronized wallet's current
+    /// external receive address.
     TransactionsChanged {
         activities: Vec<Activity>,
         transaction_details: Vec<TransactionDetails>,
         balance: WalletBalance,
         tx_count: u32,
         block_height: u32,
         account_type: AccountType,
+        next_unused_external_address: AddressInfo,
```
ovitrif: I see the UniFFI `WatcherEvent::TransactionsChanged` wire layout gaining `next_unused_external_address` here, but the tracked Python package was not regenerated: `bindings/python/bitkitcore/bitkitcore.py` still reads only through `account_type`, `bindings/python/setup.py` still reports `0.5.9`, and the bundled `libbitkitcore.dylib` is byte-for-byte unchanged. This leaves the Python bindings without the new field; if the stale converter is paired with the new core, `_UniffiConverterRustBuffer.lift` rejects the unread `AddressInfo` bytes as junk and drops every transaction-change callback. Could we regenerate and commit the Python wrapper, package version, and native library from this head?

src/modules/onchain/listener.rs
```diff
@@ -382,6 +384,14 @@ fn build_tx_changed_event(
     }
 
     let tx_count = u32::try_from(activities.len()).unwrap_or(u32::MAX);
+    let next_unused_external_address = match next_unused_external_address(wallet, base_path) {
+        Ok(address) => address,
+        Err(error) => {
```
ovitrif: I see the new `Err` branch from `next_unused_external_address` being part of the watcher contract, but both added tests construct wallets whose `get_address(AddressIndex::LastUnused)` succeeds and only assert `TransactionsChanged`. A regression that panics or drops the event on derivation failure would therefore pass this coverage even though callers rely on `WatcherEvent::Error`. Could we add a controllable failing address lookup and assert that `build_tx_changed_event` returns the expected `WatcherEvent::Error`?

ovitrif (commented):
The checked-in Python binding is stale relative to the watcher payload, and the new derivation-error branch seems to lack focused test coverage. Please check my inline comments.

src/modules/onchain/listener.rs
ben-kaufman: Fixed in c1a87a2. I added a deterministic test using a valid bare `pk(...)` descriptor that makes BDK return `ScriptDoesntHaveAddressForm`, then asserts that `build_tx_changed_event` returns `WatcherEvent::Error` with the expected message.

src/modules/onchain/types.rs
ben-kaufman: Fixed in c1a87a2. I regenerated the Python wrapper, setup version, and bundled native library with `./build_python.sh`. I also verified the package imports and the watcher event constructor now includes `next_unused_external_address`.

ovitrif:
Thanks — I’ve rebased #145 onto this branch; after #147 merged, GitHub retargeted #145 to `master`.

The test fix itself did not require another binding regeneration or version bump here. #145 owns the follow-up [`v0.5.11`](https://github.com/synonymdev/bitkit-core/releases/tag/v0.5.11) release with regenerated iOS, Android, and Python artifacts, while preserving the `v0.5.10` changelog entry for this watcher change.

ovitrif (approved):
utACK

PR #152, opened 2026-09-03
feat: Blockstream Jade hardware wallet support

coreyphillips:
Adds Blockstream Jade as a third hardware wallet, alongside Trezor and Foundation Passport.

The Jade protocol lives in a new crate, [jade-client-rs](https://github.com/coreyphillips/jade-client-rs), and this PR carries the UniFFI adapter that exposes it to the apps. There is no existing Rust client for Jade, so the CBOR protocol was implemented from the firmware and `jadepy` sources.

> Draft: the protocol has not yet been exercised against physical hardware. See QA Notes.

### Description

Scope is Bitcoin single signature, matching what Bitkit does today. Liquid, multisig, firmware updates and the airgapped QR mode are out of scope.

**FFI surface** (18 functions): scan, connect, disconnect, cancel, ping, unlock, logout, version info, xpub, account export, master fingerprint, address verification, message signing, PSBT signing, plus the transport callback setter.

Jade returns a **signed PSBT**, so it follows the Passport route rather than the Trezor one:

```
onchain_compose_transaction -> jade_sign_psbt -> finalize_psbt -> onchain_broadcast_raw_tx
```

**Transports.** Bluetooth on every platform, driven by the app through a `JadeTransportCallback` in the same shape as `TrezorTransportCallback`. USB CDC serial additionally on desktop and Python, driven from Rust.

**Why a separate crate.** The protocol is useful outside Bitkit, iterating on it does not need a bitkit-core release, and it keeps `ciborium`, `serde_bytes` and `serialport` out of this repo's direct dependencies. The split follows the `trezor-connect-rs` precedent, with two improvements worth noting for review:

- Types use `#[uniffi::remote]` rather than mirrored structs. That generates the same scaffolding a derive would, against types defined in another crate, so there are no parallel definitions and no hand-written `From` conversions in either direction. The trezor module carries roughly 900 lines of exactly that; the adapter here is 969 lines total.
- Transport failures cross the boundary as a typed `JadeTransportErrorCode`. The trezor adapter has to encode its code into a sentinel string and parse it back out, because its upstream crate offers no typed channel.

**Things a reviewer should look at deliberately:**

- **The dependency is a git rev, not a crates.io version.** Deliberate, so this never depends on an unpublished release. It should move to a version before this merges, or the merge should accept the git pin knowingly.
- **`HardwareWalletVendor::Blockstream` is a new enum case.** That makes exhaustive Kotlin `when` and Swift `switch` over the vendor enum non-exhaustive, which is source breaking for consuming apps. Appended after `Foundation`, since UniFFI assigns discriminants by declaration order.
- **Bindings are not regenerated** in this PR, and there is no version bump.
- **No `u8` or `u16` in the FFI surface.** `ping` returns an enum and `battery_status` is `u32`, keeping this module off the narrow unsigned return path that 0.5.14 fixed for Android ARM32.

Three protocol details produce code that compiles and then fails only against hardware. Each is verified against firmware source and covered by a test in the crate:

- Binary fields must be CBOR byte strings. serde encodes a plain `Vec<u8>` as an array of integers, which the device rejects for `psbt` and `entropy`.
- Replies with id `"00"` are terminal errors, not stray frames. Jade uses that id when it rejects a message before recovering the real one, so discarding them turns every such rejection into a full length timeout.
- An HTTP failure during unlock must still send `pin` with no params, or the device stays blocked and consumes the next unrelated request as the awaited reply.

### Preview

Not applicable; this is FFI surface with no UI. `src/modules/jade/README.md` documents the architecture, and the crate's README documents the wire protocol and the Bluetooth contract a native transport must honour.

### QA Notes

**Not yet run against a physical Jade.** That is the main thing this draft is waiting on. Two constants need confirming on hardware: `MIN_JADE_FIRMWARE`, and the `m/0'` parent-fingerprint route used for the master fingerprint.

Automated:

```bash
cargo clippy --all-targets -- -D warnings
cargo test modules::jade              # 6 adapter tests
cargo test modules::hardware_wallet   # catalog, count bumped 6 -> 8
cargo test -- --skip modules::blocktank
```

473 pass, 0 fail. Blocktank is skipped because its tests reach `api.stag.blocktank.to` and fail on any machine without access to it; that is pre-existing on `master`.

Protocol level coverage lives in the crate (`cargo test` there, 54 tests against a scripted mock device and a fake pinserver, no hardware or network needed).

Platform gating, which matters because `serialport` must never reach a mobile build:

```bash
cargo check --target aarch64-apple-ios      # serialport absent
cargo check --target aarch64-linux-android  # serialport absent; needs NDK clang on PATH
```

Hardware path once a device is available, over USB so no app is required:

```bash
./build_python.sh
```

then scan, connect, `jade_unlock(Testnet)` entering the PIN on device, master fingerprint returns 8 hex chars, `jade_get_xpub("m/84'/1'/0'")` returns a tpub, `jade_verify_address` matches a host-derived address, `jade_sign_message` verifies against the returned address, and `onchain_compose_transaction` to `jade_sign_psbt` to `finalize_psbt` yields a stable txid. Broadcast on regtest or testnet only.

Bluetooth end to end needs an app-side `JadeTransportCallback`. The contract is documented on the trait; the two second inter-chunk deadline and the write-with-response requirement are the parts that fail silently if missed.


coreyphillips:
Superseded by #153. The branch was renamed from `feat/jade-via-crate` to `feat/jade-hardware-wallet`, which closed this PR rather than retargeting it. Same commits, same diff.

PR #153, opened 2026-09-03
feat: Blockstream Jade hardware wallet support

coreyphillips:
Adds Blockstream Jade as a third hardware wallet, alongside Trezor and Foundation Passport.

The Jade protocol lives in [jade-client-rs](https://github.com/synonymdev/jade-client-rs). This PR carries the UniFFI adapter that exposes it to the apps. There is no existing Rust client for Jade, so the CBOR protocol was implemented from the firmware and `jadepy` sources.

### Description

Scope is Bitcoin single signature, matching what Bitkit does today. Liquid, multisig, firmware updates and the airgapped QR mode are out of scope.

**FFI surface** (18 functions): scan, connect, disconnect, cancel, ping, unlock, logout, version info, xpub, account export, master fingerprint, address verification, message signing, PSBT signing, plus the transport callback setter.

Jade returns a **signed PSBT**, so it follows the Passport route rather than the Trezor one:

```
onchain_compose_transaction -> jade_sign_psbt -> finalize_psbt -> onchain_broadcast_raw_tx
```

**Transports.** Bluetooth on every platform, driven by the app through a `JadeTransportCallback` in the same shape as `TrezorTransportCallback`. USB CDC serial additionally on desktop and Python, driven from Rust.

**Why a separate crate.** The protocol is useful outside Bitkit, iterating on it does not need a bitkit-core release, and it keeps `ciborium`, `serde_bytes` and `serialport` out of this repo's direct dependencies. Compared with the `trezor-connect-rs` split:

- Types use `#[uniffi::remote]` rather than mirrored structs, so there are no parallel definitions and no hand-written `From` conversions.
- Transport failures cross the boundary as a typed `JadeTransportErrorCode` instead of a sentinel string.

**Worth reviewing deliberately:**

- **The dependency is a git rev, not a crates.io version.** Pinned at `d52ccd9`. It moves to a published version once the crate is on crates.io.
- **`HardwareWalletVendor::Blockstream` is a new enum case.** Exhaustive Kotlin `when` and Swift `switch` over the vendor enum will need a new branch, so this is source breaking for consuming apps. Appended after `Foundation`, since UniFFI assigns discriminants by declaration order.
- **No `u8` or `u16` in the FFI surface**, keeping this module off the narrow unsigned return path that 0.5.14 fixed for Android ARM32.
- Bindings are regenerated and the version is bumped to 0.5.16.
- **`gradle-publish.yml`** now installs only `platform-tools` in the Android SDK step, because `setup-android` failed trying to install the removed `tools` package. A manual run also uploads the native debug symbols when a matching release exists.
- **`finalize_psbt`** now rejects any signature that does not commit to the whole transaction (ECDSA other than `All`, Schnorr other than `Default` or `All`). This also applies to Passport, which uses the same path.

### Preview

Not applicable; this is FFI surface with no UI. `src/modules/jade/README.md` documents the architecture, and the crate's README documents the wire protocol and the Bluetooth contract.

### QA Notes

**Run against a physical Jade v1**, firmware 1.0.41, from macOS, over both transports.

| | Serial | Bluetooth |
|---|---|---|
| Connect, version info, ping | yes | yes |
| Unlock through the pinserver | yes | yes |
| Account export | yes | yes |
| Address verification, `wpkh` and `tr` | yes | `wpkh` |
| Message signing | yes | yes |
| PSBT signing | yes | yes |
| Fragmented reply via `get_extended_data` | not reached | yes |
| Cancellation, logout | yes | not reached |

Signatures were verified against the keys the account xpub derives, and every PSBT spent a fabricated prevout, so nothing was broadcastable. A 20 input PSBT exercised chunked Bluetooth writes and a fragmented reply reassembled through `get_extended_data`.

- The `m/0'` parent fingerprint route is confirmed: the device signed a PSBT whose BIP32 origins carried that fingerprint.
- `MIN_JADE_FIRMWARE` (`0.1.48`) is still unconfirmed. Only 1.0.41 has been exercised; the device answers `UNKNOWN_METHOD` for anything it cannot do.
- Untested: Jade Plus and Linux serial.

Automated:

```bash
cargo clippy --all-targets -- -D warnings
cargo test modules::jade
cargo test modules::hardware_wallet
cargo test -- --skip modules::blocktank
```

Blocktank is skipped because its tests reach `api.stag.blocktank.to`; that is pre-existing on `master`. Protocol level coverage lives in the crate (72 tests against a scripted mock device and a fake pinserver).

Platform gating, since `serialport` must never reach a mobile build:

```bash
cargo check --target aarch64-apple-ios
cargo check --target aarch64-linux-android
```

Bluetooth end to end needs an app-side `JadeTransportCallback`. Reference material for the Swift and Kotlin sides:

- [`examples/callback_transport.rs`](https://github.com/synonymdev/jade-client-rs/blob/main/examples/callback_transport.rs) runs the callback shape against a scripted link with no hardware.
- [`docs/bluetooth.md`](https://github.com/synonymdev/jade-client-rs/blob/main/docs/bluetooth.md) covers service UUIDs, chunk sizing per platform, and the rules that fail silently when missed.


src/modules/jade/implementation.rs
```diff
@@ -0,0 +1,424 @@
...
+            .decode(psbt.trim())
+            .map_err(|error| JadeError::InvalidPsbt {
+                error_details: format!("base64 decoding failed: {error}"),
+            })?;
+        let parsed = Psbt::deserialize(&bytes).map_err(|error| JadeError::InvalidPsbt {
+            error_details: format!("parsing failed: {error}"),
+        })?;
+
+        let mut guard = self.session.lock().await;
+        let signed = guard
```
jvsena42: **A returned `tap_key_sig` is never checked for its sighash flag, so a key-holding device can hand back a signature that lets the outputs be rewritten.**

`verify_signed_psbt` (jade-client-rs `client.rs:606-616`) only checks that a signature was *gained* — `tap_key_sig.is_some() && before.tap_key_sig.is_none()`. The flag byte rides inside the signature, so `check_signable` (`client.rs:486-505`), which inspects the *sent* PSBT's `input.sighash_type`, does not constrain what comes back.

Downstream does not close it either. `finalize_psbt` (`src/modules/onchain/psbt.rs:11-49`) checks `unsigned_tx` and per-input prevout equality but no flags. miniscript 12.3.7 `sanity_check` (`src/psbt/mod.rs:380-420`) iterates `partial_sigs` only, so it catches ECDSA and never looks at `tap_key_sig`; the interpreter's Schnorr branch (`src/interpreter/mod.rs:250-255`) computes the sighash using the type embedded in the signature, so a `0x82` key-spend signature verifies on its own terms.

Failure scenario: compromised Jade firmware answers a P2TR `sign_psbt` with a 65-byte `tap_key_sig` ending `0x82` (`NONE|ANYONECANPAY`). Verification passes, `finalize_psbt` finalizes, the app broadcasts, and anyone watching the mempool can rewrite the outputs.

Pre-existing, and not a blocker for this PR: `finalize_psbt` arrived with Passport in #138 and is already on `master`; this PR adds a second signer feeding the same path. The threat model needs firmware that already holds the seed, so treat it as defense in depth against one specific vector rather than a new hole.

Narrowest fix covers both signers, in `finalize_psbt`: reject a `tap_key_sig` whose `sighash_type` is not `TapSighashType::Default`, and for symmetry a `partial_sigs` flag other than `All`. Reasonable as a follow-up rather than here.

src/modules/jade/implementation.rs
```diff
@@ -0,0 +1,424 @@
...
+        let matches = self
+            .connected_device
+            .read()
+            .await
+            .as_ref()
+            .map(|device| device.path == path)
+            .unwrap_or(false);
+        if matches {
+            log::debug!("[jade] native layer reported a disconnect");
+            self.connection_changes.send_replace(());
```
jvsena42: **This broadcast can only reach a queued `connect`, and it makes that reconnect fail with `UserCancelled` when nothing was cancelled.**

`notify_disconnected` already holds `lifecycle` (line 266), so no handshake can be running while it fires. `connect` is the only subscriber (line 135) — `disconnect` and `cancel` never subscribe. A `connect` parked on the `lifecycle` mutex at line 136 subscribed before this `send_replace`, so `has_changed()` at line 137 is true when it wakes and it returns `Err(UserCancelled)` at 138.

Failure scenario: the app handles one BLE `DISCONNECTED` event by calling `jade_notify_disconnected(path)` and `jade_connect(path)` to auto-reconnect without sequencing them. If notify takes `lifecycle` first, the user sees "cancelled" on a reconnect they asked for. `jade_notify_disconnected` does await its `JoinHandle` (`lib.rs:2705-2707`), so a caller that awaits it before reconnecting never hits this — it needs a caller that does not.

Teardown does not depend on this line: `disconnect_session` (237-248) closes the transport through the `CancelHandle` and waits on the `session` mutex; the watch channel plays no part.

Deleting line 276 fixes that direction but leaves the mirror: if the reconnect wins the lock instead, this call then tears down the freshly opened same-path session at line 277. If a generation token is more than you want here, one sentence on `jade_notify_disconnected` (`lib.rs:2699-2701`) stating it must complete before a reconnect for the same path is issued covers both directions.

src/modules/jade/implementation.rs
```diff
@@ -0,0 +1,424 @@
...
+        }
+        let connecting = Jade::connect(Arc::clone(&transport));
+        tokio::pin!(connecting);
+        let session = tokio::select! {
+            biased;
+            _ = changes.changed() => {
+                let closed = transport.close().await;
+                // Native callbacks already running on the blocking pool must finish
+                // before a replacement can reuse the same device path.
+                let _ = connecting.await;
```
jvsena42: **Disconnecting during a handshake can block for up to 60s, because nothing forces the pending read to fail.**

This arm closes the transport (line 175) then awaits the handshake. `Jade::connect` has no abort flag reachable from here — the `aborted` flag is set through a `CancelHandle`, and this path uses `CallbackTransport::close` instead. The read loop (jade-client-rs `transport.rs:266-293`) then exits only on a read error, a matched reply, or the 60s `QUICK_TIMEOUT` (`client.rs:32`), sleeping 25ms per empty chunk.

`read_some` (`callbacks.rs:197-213`) returns whatever the foreign `read_chunk` gives it, and `callbacks.rs:91` tells implementers that success with an empty vector means "nothing yet". So a BLE implementation that drains a notification queue and keeps reporting success after `close_device` is contract-compliant, and `jade_disconnect` then holds `lifecycle` for the full 60s with any `jade_connect` queued behind it.

The lifecycle tests do not catch this because the mock returns `Disconnected` for a closed path.

This needs no code change if the contract says so: add a fourth requirement to the list at `callbacks.rs:64-74` — `close_device` must make a concurrent `read_chunk` on the same path return an error. Worth pinning down now, while the Kotlin and Swift transports are still being written against it.

src/modules/jade/implementation.rs
```diff
@@ -0,0 +1,424 @@
...
+            .map_err(|_| JadeError::DeviceBusy)?;
+        // Starting a Bluetooth scan while a GATT link is up reliably drops it on
+        // Android, so refuse rather than silently breaking the open session.
+        if self.connected.load(Ordering::SeqCst) {
+            return Err(JadeError::DeviceBusy);
+        }
+
+        let mut discovered: Vec<JadeDeviceInfo> = Vec::new();
+
+        if let Some(callback) = transport_callback() {
```
jvsena42: **With no transport callback registered, `jade_scan` returns `Ok([])` on iOS and Android instead of `NotInitialized`.**

Without a callback this block is skipped, line 98's serial enumeration is `cfg`-gated off on mobile, and line 101 stores an empty list — so the call succeeds reporting no devices. `build_transport` treats the same missing callback as `NotInitialized` (line 210), and the Trezor sibling returns it from `scan` too (`trezor/implementation.rs:445-446`).

Failure scenario: `jade_set_transport_callback` is missed after an activity restart or a wiring-order bug. The picker shows "No devices found" indefinitely, with no typed error for the app to distinguish from a genuinely empty scan.

Fix: under `cfg(any(target_os = "ios", target_os = "android"))`, `.ok_or(JadeError::NotInitialized)?` here. Desktop legitimately has no callback and relies on serial enumeration, so it has to stay gated.

src/modules/jade/implementation.rs
```diff
@@ -0,0 +1,424 @@
...
+    }
+
+    pub async fn connected_device(&self) -> Option<JadeDeviceInfo> {
+        self.connected_device.read().await.clone()
+    }
+
+    /// The version summary read at connect, or refreshed since.
+    pub async fn version_info(&self) -> Option<JadeVersionInfo> {
+        self.session
+            .lock()
```
jvsena42: **`jade_get_version_info` takes the session mutex just to clone a cached struct, so it blocks behind whatever operation is in flight.**

`sign_psbt` (416), `verify_address` (379), `sign_message` (393) and `unlock` (315) all hold `self.session.lock()` across their device awaits, bounded by `CONFIRM_TIMEOUT` / `UNLOCK_TIMEOUT` of 300s. A signing screen that reads version info for battery or state while the user is confirming on the device therefore suspends for up to five minutes, and since UniFFI async exports are detached, cancelling the foreign task does not release it. `jade_is_connected` and `jade_get_connected_device` avoid this.

`lib.rs:2723` describes it as "read at connect, without touching the device" — true about I/O, misleading about latency.

Narrowest fix is `try_lock`, returning `None` when an operation is in flight so the app can pair it with `jade_is_connected`; amending that doc line would do as well. A separately mirrored copy of the struct is more machinery than this needs.

jvsena42 (commented):
Reviewed this as a funds-and-key change: PSBT signing, xpub and account export, address verification, and a CBOR parse boundary fed by the device. Read the adapter and `jade-client-rs` at the pinned rev `d52ccd9` in full.

No blocking findings. Five LOW notes inline, none of which need to hold the PR.

**Fund safety is the part I most wanted to break, and it holds.** `sign_psbt` never trusts the device's reply: `verify_signed_psbt` (`client.rs:568-617`) compares `unsigned_tx` — so every output, amount and input outpoint — plus input/output counts, and `witness_utxo` / `non_witness_utxo` txid when echoed, and requires at least one new signature. The adapter (`implementation.rs:406-423`) only decodes, deserializes and re-encodes that verified result, and `finalize_psbt` re-checks `unsigned_tx` and prevouts again. The one gap is the sighash flag on a returned taproot signature, noted inline and pre-existing.

### Checked and clean

- **Key material.** No xpub, fingerprint, derivation path, PIN or PSBT bytes reach a log. Adapter logging is three statements, two of them constant text. Crate logs carry method names, ids and byte counts, and reqwest errors go through `without_url()`. `#![deny(clippy::print_stdout, print_stderr, dbg_macro)]`. Nothing written to disk.
- **Pinserver.** The URL is device-supplied and validated: https only, no credentials, port 443 only, no `.onion`, DNS resolved and pinned to a single public address with private/CGNAT/link-local/v4-mapped rejected, redirects disabled, body capped at 64 KiB, at most 4 round trips, `on-reply` must be `pin`. No `danger_accept_invalid_certs` and no http fallback anywhere in the crate. A non-default host only warns rather than failing, which is deliberate and documented; the host is a courier for a device-encrypted blob either way.
- **Protocol bounds.** Frames capped at 64 KiB, reassembly capped at 64 fragments and 64 KiB with `checked_add`, fragment sequence echoed and checked, absolute deadline, buffer cleared on malformed CBOR, connection poisoned on violation. Derivation paths capped at depth 8. No `unwrap`/`expect`/`panic!` in crate or adapter source.
- **Address verification.** A purpose/variant mismatch is rejected and the device's returned address is compared against the expected one, so a wrong path cannot report success unless the host derived the expectation from that same wrong path.
- **Lock discipline.** `connect` closes any prior session first, checks the watch before and after the native open, and cancels the handshake via `select`. The cancel handle lives outside the session lock, so `disconnect` returns promptly during an in-flight operation. Lock order is consistent between `connect` and `disconnect_session`. Separate manager and callback statics per vendor, with no shared fingerprint, xpub or session cache.
- **Your flagged items all hold.** `Blockstream` is last in the Rust enum and last in both generated bindings, with `Trezor`=1 and `Foundation`=2 unchanged. No `u8`/`u16` in any of the 20 `jade_*` exports. `0.5.16` is consistent across `Cargo.toml`, `Package.swift`, `gradle.properties` and `setup.py`, and the `Package.swift` checksum matches the checked-in xcframework zip. `Cargo.lock` pins `d52ccd9099f28aa75fc30f4b94315c2afb64321c`. `cargo tree` for `aarch64-linux-android` and `aarch64-apple-ios` shows no `serialport`.

### One note on the release, not the code

The description presents the new `HardwareWalletVendor` case as source-breaking for consuming apps, which reads as upcoming. It already shipped. `v0.5.15` (2026-09-07) and `v0.5.16` (2026-09-15, currently Latest) are both published releases whose tagged commits carry `src/modules/jade` and are ancestors of this branch but not of `master`. Consumers therefore already have the new case, and the current Latest release contains code that is not on `master`. Verified with `git cat-file -p "<tag>^{commit}:<path>"`, since `git show <tag>:<path>` misreports on these annotated tags.

Consumer side: https://github.com/synonymdev/bitkit-android/pull/1231 is the app-side PR for this FFI; there is no iOS counterpart open.

### Not verified

I did not execute the test suite — a `protoc` dependency in the `trezor-connect-rs` build script and then disk pressure stopped two attempts. `gh pr checks 153` reports no checks at all on this head, so as far as I can see nothing has run `cargo test` against it. The checked-in iOS and Python binaries cannot be inspected; the checksum match proves `Package.swift` references that zip, not that the zip was built from this source.

.github/workflows/gradle-publish.yml
```diff
@@ -80,10 +82,9 @@ jobs:
         path: bindings/android/native-debug-symbols.zip
 
     - name: Upload native debug symbols to release
-      if: github.event_name == 'release'
       env:
         GH_TOKEN: ${{ github.token }}
-      run: gh release upload "${{ github.event.release.tag_name }}" bindings/android/native-debug-symbols.zip --clobber
+      run: gh release upload "${{ github.event.release.tag_name || inputs.version }}" bindings/android/native-debug-symbols.zip --clobber
```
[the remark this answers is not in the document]
coreyphillips: Fixed in 74efa46. The step still runs on a manual dispatch, because that is how the v0.5.16 symbols got attached after the release event failed, but it now checks `gh release view "$TAG"` first and skips the upload when no release exists, so a manual publish always reaches `Publish to GitHub Packages`.

src/modules/jade/README.md
```diff
@@ -0,0 +1,133 @@
...
+a scripted mock device and a fake pinserver.
+
+## Before device testing and release
+
+The checked-in iOS and Python binaries are intentionally awaiting a rebuild after
+review feedback is addressed. They currently contain an older Jade client than
+the source dependency pin. Before testing the final changes on live devices,
+rebuild the platform artifacts from the final source revision, update the
+XCFramework checksum in `Package.swift`, and verify the regenerated bindings and
+native libraries together. The version remains 0.5.15 during this preparation.
```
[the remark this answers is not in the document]
coreyphillips: Fixed in 74efa46. Removed the `Before device testing and release` section.

src/modules/jade/implementation.rs
```diff
@@ -0,0 +1,424 @@
...
+            .sign_message(network, &derivation_path, &message)
+            .await
+    }
+
+    /// Sign a base64 PSBT and return the signed PSBT, base64 encoded.
+    ///
+    /// The FFI surface speaks base64 because that is what `compose_transaction`
+    /// emits and what `finalize_psbt` expects; the protocol crate works in typed
+    /// PSBTs, so the encoding boundary lives here.
+    pub async fn sign_psbt(&self, network: JadeNetwork, psbt: String) -> Result<String, JadeError> {
```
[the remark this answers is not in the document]
coreyphillips: Fixed in 74efa46. Added `malformed_base64_psbt_is_rejected_before_reaching_the_device` (asserts `InvalidPsbt`) and `native_disconnect_clears_only_the_connected_path` (a notification for another path leaves the session, one for the connected path clears the session, device and version info, and closes the native handle).

src/modules/jade/implementation.rs
coreyphillips: Fixed here rather than as a follow-up, since it covers both signers. In 74efa46 `finalize_psbt` runs the miniscript interpreter over the finalized satisfaction after `interpreter_check` and rejects any ECDSA signature whose flag is not `All`, and any Schnorr signature that is not `Default` or `All`. Walking the finalized satisfaction instead of `partial_sigs` / `tap_key_sig` also covers a signer that returns inputs already finalized, which the field level check would miss. Tests: `rejects_ecdsa_signature_that_does_not_commit_to_outputs` (SIGHASH_NONE), `rejects_taproot_signature_that_does_not_commit_to_outputs` (NONE|ANYONECANPAY key spend), and `finalizes_taproot_key_spend_psbt` for the valid case.

src/modules/jade/implementation.rs
coreyphillips: Fixed in 74efa46. Removed the `send_replace` from `notify_disconnected`, with a comment on why it is unnecessary while `lifecycle` is held, so a reconnect queued behind the notification proceeds instead of failing with `UserCancelled`. For the mirror case, `jade_notify_disconnected` now documents that it must be awaited before reconnecting the same path.

src/modules/jade/implementation.rs
coreyphillips: Fixed in Rust rather than only in the contract, in 74efa46. `CallbackTransport::close` sets a closed flag, and `read_some` / `write_all` (including between chunks) return `Disconnected` once it is set, so the handshake read loop exits on its next iteration no matter what the native `read_chunk` keeps returning. The trait docs note the behaviour. `disconnect_does_not_wait_for_a_native_layer_that_keeps_reading_after_close` uses a mock that keeps reporting empty successful reads after close and asserts `disconnect` returns within 1s; it fails without the flag.

src/modules/jade/implementation.rs
coreyphillips: Fixed in 74efa46. On iOS and Android `scan` now returns `NotInitialized` when no transport callback is registered; desktop keeps the optional callback plus serial enumeration. Checked with `cargo check --target aarch64-apple-ios`.

src/modules/jade/implementation.rs
coreyphillips: Fixed in 74efa46. Went with a cached copy after all instead of `try_lock`, because returning `None` while connected would read as "not connected" to the apps. `version_info` is a `RwLock<Option<JadeVersionInfo>>` beside `connected_device`, set at connect and by `refresh_version_info`, and cleared in `disconnect_session`, so `jade_get_version_info` never touches `session`. Doc line updated. `version_info_does_not_wait_for_an_operation_in_flight` holds a ping open and reads version info within 200ms; it fails with the old implementation.

ovitrif:
Is this PR also integrating the Jade simulator/ emulator? Or it is more HTTP-based, thus a better fit for `bitkit-docker`?

coreyphillips:
No, this PR doesn't integrate the emulator, and the emulator isn't HTTP either.

Blockstream's jade-emulator is a QEMU ESP32 image that speaks the same CBOR byte-stream protocol as the real device, exposed over a raw TCP socket (`tcp:localhost:30121`). So it's effectively a third transport alongside the two here (Bluetooth via the foreign callback, USB CDC serial on desktop), not a web service.

The only HTTP in the Jade path is the pinserver, and that's Blockstream's hosted blind PIN oracle reached over HTTPS from `jade-client-rs` (`reqwest-pinserver` feature). It's remote, so it doesn't need a container.

If we do want emulator coverage later, the split would be:

- a `TcpTransport` in `jade-client-rs`, where the protocol and its tests already live. It can't go in bitkit-core: bindings are generated from the host library, so a desktop-only `#[uniffi::export]` would show up in the generated Swift and Kotlin while being missing from the device library.
- the `blockstream/jade-emulator` container in `bitkit-docker` as a dev service, which is the part your instinct is right about.

Worth noting it wouldn't cover much of what's risky in this PR: the emulator is serial/TCP only, so the BLE contract (the 2s inter-chunk deadline, the write-with-response requirement) still only fails against real hardware. Today the adapter is covered by `cargo test modules::jade`, and the protocol level is covered in `jade-client-rs` against a scripted mock device and a fake pinserver.

On Passport: our support for it is UR/QR only (`src/modules/ur/passport.rs`), airgapped with no transport at all, so its simulator sits at a different layer than anything this PR touches.

jvsena42:
@coreyphillips conflicts

.github/workflows/gradle-publish.yml
```diff
@@ -80,10 +82,18 @@ jobs:
...
 
     - name: Upload native debug symbols to release
-      if: github.event_name == 'release'
       env:
         GH_TOKEN: ${{ github.token }}
-      run: gh release upload "${{ github.event.release.tag_name }}" bindings/android/native-debug-symbols.zip --clobber
+        TAG: ${{ github.event.release.tag_name || inputs.version }}
+      shell: bash
+      run: |
+        # A manual publish may target a version with no GitHub release; skip the
```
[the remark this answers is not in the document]
coreyphillips: Fixed in a8e0b99. A `release` event now uploads unconditionally, since a release definitionally exists there and any failure should fail the step. A `workflow_dispatch` still does the lookup, but it captures stderr and only takes the skip branch when the message says the release was not found; anything else (auth, network, rate limiting) prints the error and exits 1, so a publish without symbols can no longer pass silently.

src/modules/jade/callbacks.rs
```diff
@@ -0,0 +1,253 @@
...
+            Ok(())
+        })
+        .await
+        .map_err(|error| JadeError::IoError {
+            error_details: format!("write task failed: {error}"),
+        })?
+    }
+
+    async fn read_some(&self, timeout: Duration) -> Result<Vec<u8>, JadeError> {
+        self.ensure_open()?;
```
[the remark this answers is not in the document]
coreyphillips: Fixed in a8e0b99, docs plus a small code change.

The trait docs now say that a `read_chunk` already inside the native layer cannot be interrupted, so requirement 3 is what bounds a disconnect issued mid-handshake. Requirement 3 also names the number now: `read_some` is called with at most `READ_CHUNK_TIMEOUT_MS` (250ms) from the crate's read loop, so an implementation that honours `timeout_ms` delays a disconnect by that much at worst.

On the code side, `read_some` re-checks the closed flag after the callback returns and yields `Disconnected` instead of the data, so bytes that arrive for a released path are dropped rather than fed back to the parser. `bytes_arriving_after_close_are_discarded` parks a mock inside `read_chunk`, closes while it is parked, then releases it; it fails without the post-call check.

coreyphillips:
Conflicts resolved in 76eebaf, `master` merged in.

Every conflict was a modify/delete on generated binding artifacts: this branch regenerated them for 0.5.16, and #154 stopped tracking them and added them to `.gitignore`. Took master's deletion in all eight cases, so the generated Kotlin, the xcframework and the Python package are no longer checked in here either. The new bindgen workflows regenerate them.

Nothing in `src/` or the build scripts conflicted, and `gradle-publish.yml` kept this branch's version since master hasn't touched that file since the merge base. Versions are still in sync at 0.5.16 across `Cargo.toml` and `gradle.properties`, with `Package.swift` now on master's tag interpolation.

Checked against what Bindgen Validation runs: `bash -n` on the four build scripts, `cargo metadata --locked`, and the workflow YAML all pass. `cargo test modules::jade` is 19/19 and `cargo test modules::activity` is 196/196.

jvsena42 (commented):
Re-reviewed the delta since 43c3ffd: 74efa46, a8e0b99, the master merge 76eebaf and the rustfmt commit 9a53bf3. I read it with jade-client-rs at `d52ccd9`, miniscript 12.3.7 and the consumer side on https://github.com/synonymdev/bitkit-android/pull/1231.

No blocking findings. One LOW note inline, on a doc promise the crate does not keep.

### Earlier threads, checked at head

- **Sighash flags in `finalize_psbt`: holds.** `validate_signature_hash_types` runs after `interpreter_check` over every input and reads `final_script_sig` / `final_script_witness`, so inputs a signer returns already finalized are covered too. Signatures are 64-byte (Default) or 65-byte with `0x01`, and both are accepted. Key-spend, P2PKH/P2WPKH/P2SH-P2WPKH and script paths including `multi_a` all produce a signature constraint, so none is skipped. The walk re-evaluates the same satisfaction `interpreter_check` already verified, so it cannot newly reject a PSBT that Passport or Jade legitimately signs.
- **`send_replace` in `notify_disconnected`: holds.** It is gone, and `jade_notify_disconnected` documents the await-before-reconnect rule.
- **Disconnect during the handshake: holds.** `closed` belongs to each `CallbackTransport`, and `build_transport` creates a fresh one on every connect, so reconnecting to the same path cannot inherit a stale flag.
- **`jade_scan` on mobile without a callback: holds.** It returns `NotInitialized` without touching the previous device list.
- **`jade_get_version_info`: holds.** It reads only the `RwLock`.
- **Bot threads (workflow guard, README, adapter tests, in-flight `read_chunk` docs): all hold.**

### Also clean

- The remerge diff of 76eebaf is exactly the eight generated-artifact deletions. Nothing under `src/`, `Cargo.*` or `Package.swift` came from conflict resolution.
- The delta adds no logging of PSBT bytes, xpubs, paths or version details.
- No FFI signature changed between the `v0.5.16` tag and head. The only semantic changes (`jade_scan`, `jade_get_version_info`, `finalize_psbt`) all favour the consumer.
- `finalize_psbt` has no caller on bitkit-android or bitkit-ios `master` yet. The stricter sighash check first reaches users with the Jade app PR.

### Not verified

I did not run `cargo test` because `protoc` is missing locally. CI's `Android Bindgen` and `iOS Bindgen` were still pending on this head when I checked.

src/lib.rs
```diff
@@ -2598,8 +2606,318 @@ pub async fn trezor_clear_credentials(device_id: String) -> Result<(), TrezorErr
...
+            Err(JadeError::IoError {
+                error_details: format!("Runtime error: {}", e),
+            })
+        })
+}
+
+/// Close the device and clear session state.
+///
+/// Safe to call while an operation is waiting on a confirmation: the pending
+/// request returns `UserCancelled` promptly rather than running out its deadline.
```
jvsena42: **Most of the time, a disconnect during a confirmation wait returns `DeviceDisconnected`, not the `UserCancelled` this doc promises.**

`disconnect_session` → `CancelHandle::cancel` sets `aborted`, then closes the transport. The crate's reply loop (jade-client-rs `transport.rs:266-269`) checks `aborted` only at the top of each iteration. The read arm (`:278-283`) poisons and returns the transport error as-is. Each iteration spends up to 250ms in `read_some` and 25ms in the idle sleep, so a disconnect nearly always lands mid-read. The post-callback check at `callbacks.rs:234-236` then yields `Disconnected`, which `error.rs:44` maps to `DeviceDisconnected`. `write_request` (`:204-207`) has the same gap.

No impact on the Android app today. `HwSendViewModel` / `HwReceiveViewModel` cancel the awaiting job before calling `jade_disconnect`, and `JadeTransport.readBleChunk` already errors once closed. So the variant is never observed, and with that transport this was already the outcome before 74efa46. The lifecycle tests miss it because every `UserCancelled` assertion is on `connect`, which the manager maps itself. `version_info_does_not_wait_for_an_operation_in_flight` asserts only `is_err()`.

Narrowest fix, pick one:
- Doc only: "the pending request fails promptly (`UserCancelled` or `DeviceDisconnected`)".
- In jade-client-rs, on both error arms: `return Err(if self.aborted.load(Ordering::SeqCst) { JadeError::UserCancelled } else { error })`, plus a pin bump.

A remap inside bitkit-core would need extra state, because `disconnect_session` takes the cancel handle before the op returns. I wouldn't go that way.

jvsena42 (commented):
Re-reviewed 6644a83. It only touches `bindings/ios/bitkitcore.swift`. No new findings.

- **Complete.** It covers every doc comment changed since `9d1486f` (the last regeneration) on an exported item: `jade_notify_disconnected`, `jade_get_version_info` and the `JadeTransportCallback` docs. The other doc changes in `callbacks.rs` and `psbt.rs` are on private items, which have no UniFFI checksum. The two new checksums match what CI generated: `iOS Bindgen` → `Check committed iOS interface files` passed on this head.
- **No generated code changed** beyond those docstrings and the two checksum constants.

My note on `src/lib.rs:2670` (a disconnect mid-confirmation returns `DeviceDisconnected`, not `UserCancelled`) is still open. It's LOW and doesn't need to hold the PR. `Android Bindgen` was still pending when I checked.

ovitrif (approved):

jvsena42 (approved):

